PR #13, opened 2021-04-08
docs: plugin lifecycle

ankithans:
check [plugin docs](https://github.com/aerogear/charmil/blob/plugin-docs/docs/src/plugin.md) here


wtrocki (approved):

wtrocki:
I would prefer to have this outside repo for the moment as it can outdate fast but let's get it and refine it over

Issue #1, opened 2021-04-06
Create Specification for the project 


wtrocki:
What we need is

1. Specification for the CLI resigstry (Sample JSON including versions, binary locations etc.)
2. List of the required features and good haves

ankithans:
@wtrocki sample specification for the CLI registry
```json
{
    "name": "",
    "description": "",
    "version": "",
    "environment": "",
    "registry": "",
    "homepage": "",
    "bin": {},
    "private": false,
    "dependencies": {

    },
    "maintainers": {

    },
    "license": ""
}
```

ankithans:
Feats:
	- Use a CLI plugin as an extension in some CLI
	- Link multiple CLI's together
	- Registry Manager containing the list and metadata of CLI's
	- Github repo/pages based remote index servers for plugins which will contain docs, metadata of the plugin
	- Framework that will be able to read these remote index servers

wtrocki:
Pretty much, but also logic for listing and downloading extensions. Logic for executing extensions in host CLI

ankithans:
okay, I will start with a readme.

wtrocki:
@ankithans I have pushed to main branch example registry. If we add more CLIs - Kubectl, and OC (OpenShift) then we can have reliable sample app that we can base on.

ankithans:
> https://mirror.openshift.com/pub/openshift-v4/clients/

Okay, so we need this type of registry manager. Is it taking releases from GitHub?

wtrocki:
I cannot assign you both but we really need to work on this for proposal this week to send i. @namit-chandwani @ankithans

ankithans:
@wtrocki Now ATY what's left here?

namit-chandwani:
> @wtrocki Now ATY what's left here?

+1
And how should we divide the remaining work among ourselves?

ankithans:
@wtrocki 
shall we use this for Kafka?
```
 {
    "name": "kaf",
    "description": "Modern CLI for Apache Kafka, written in Go.",
    "version": "v0.1.39",
    "homepage": "https://github.com/birdayz/kaf",
    "binaries": {
      "darwin-amd64": "https://github.com/birdayz/kaf/releases/download/v0.1.39/kaf_0.1.39_Darwin_x86_64.tar.gz",
      "linux-amd64": "https://github.com/birdayz/kaf/releases/download/v0.1.39/kaf_0.1.39_Linux_x86_64.tar.gz",
[...]

PR #14, opened 2021-04-08
feat: add new CLIs to registry

namit-chandwani:
In response to: https://github.com/aerogear/charmil/issues/1#issuecomment-814744371

## Changes made
I have added OC (OpenShift CLI) and kubectl CLI to the registry.

### DIFF
--- a/registry/example/registry.json
+++ b/registry/example/registry.json
@@ -2,11 +2,38 @@
   {
     "name": "ODO",
     "description": "OpenShift Developer CLI",
-    "version": "v1",
-    "homepage": "https://www.example.com",
+    "version": "v2.0.7",
+    "homepage": "https://developers.redhat.com/products/odo/overview",
     "binaries": {
       "darwin-amd64": "https://mirror.openshift.com/pub/openshift-v4/clients/odo/v2.0.7/odo-darwin-amd64",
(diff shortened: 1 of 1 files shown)


registry/example/registry.json
```diff
@@ -2,14 +2,41 @@
...
     "tags": [],
     "license": "MIT"
+  },
+  {
+    "name": "OC",
+    "description": "OpenShift CLI",
+    "version": "v4.6",
+    "homepage": "https://docs.openshift.com/container-platform/4.6/cli_reference/openshift_cli/getting-started-cli.html",
+    "binaries": {
+      "linux": "https://mirror.openshift.com/pub/openshift-v4/clients/oc/4.6/linux/oc.tar.gz",
```
wtrocki, line 21: Not sure if we want to support zipped binaries.

wtrocki (approved):

registry/example/registry.json
namit-chandwani: I tried to get an uncompressed binary but couldn't find any.

Even the documentation states that first we need to download this zipped binary and then unzip it while installation.
(Ref: https://docs.openshift.com/container-platform/4.2/cli_reference/openshift_cli/getting-started-cli.html#cli-installing-cli-on-linux_cli-developer-commands)

Should I add any other CLI to the registry in place of this one?

wtrocki: I would leave it. Great corner case

Issue #7, opened 2021-04-07
Provide documentation to show how charmil differs from kubectl plugins 


wtrocki:
See: https://github.com/kubernetes-sigs/krew

https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/

wtrocki:
After investigation I see that we will be more generic version of krew that developers can install into their CLI (rather than this being available as part of the separate CLI).

Charmil will be also generic (as krew is specific to kubernetes ecosystem). 
Challenges for that will be authentication models and other factors that we will need to cather. 


I do think that we actually need some long running task to refine what charmil is and it isn't rather than define it what is now.  Created couple investigations to see how we can evolve this simple idea based on couple micro POC's

@namit-chandwani @ankithans

wtrocki:
Let's discuss if that models is clear or if you have any questions before we close this

ankithans:
charmil will be specific to kafka and knative?
model is like clear, as that of now

namit-chandwani:
> Let's discuss if that models is clear or if you have any questions before we close this

Yes, it's clear

Issue #15, opened 2021-04-12
Review and check suitability of every krew command


wtrocki:
```
krew is the kubectl plugin manager.
You can invoke krew through kubectl: "kubectl krew [command]..."

Usage:
  kubectl krew [command]

Available Commands:
  help        Help about any command
  index       Manage custom plugin indexes
  info        Show information about an available plugin
  install     Install kubectl plugins
  list        List installed kubectl plugins
  search      Discover kubectl plugins
  uninstall   Uninstall plugins
  update      Update the local copy of the plugin index
  upgrade     Upgrade installed plugins to newer versions
  version     Show krew version and diagnostics

Flags:
  -h, --help      help for krew
  -v, --v Level   number for the log level verbosity
  ```
  
 Learn about features. List them. Discuss with team What we want to support/what is later phase etc.

ankithans:
let me take this, I have initially gone through them.

wtrocki:
We need some form of the documentation that will be adjusted. I would prefer to have markdown based document or google doc -> whatever works for you

ankithans:
md is great

wtrocki:
So keep in mind that target here is not to document their CLI. It is to evaluate their CLI commands and provide info if we would like to adopt this command. 

What we are doing is basically specification for our commands influenced by theirs ;D

wtrocki:
we past this phase

Issue #18, opened 2021-04-12
Reasearch how we can create library of cobra commands


wtrocki:
Cobra on it's own provides top level frameworks for making commands. 
Making SDK for Cobra is not documented and requires some POC to be done

wtrocki:
This seems to be core of all POC - we need to prove that we would be able to use commands as SDK. 
I have done some experiments with handlers.
Expected result would be to have sample command used in example CLI (charmil it is current name I think)

ankithans:
This needs to be done first, after that all work will start. I want some help here to start.
I think this might help us here https://github.com/replicatedhq/krew-plugin-template

wtrocki:
+100

ankithans:
@wtrocki Can you check this out https://github.com/ankithans/mysdk
If this is what we want to achieve?
I have created the `sample` command in SDK and used it in the CLI

wtrocki:
Sure. Will do in couple hours.

ankithans:
yup, I was just trying things out. will create a PR tomorrow according to the project structure.

wtrocki:
I somehow think that you know how to do it..I started more detailed discussion on #26 so we can close this one

PR #19, opened 2021-04-12
docs: add features of commands

ankithans:
Features we can focus on first https://github.com/aerogear/charmil/blob/suitable-cmds/docs/src/suitable_commands.md

#15

### DIFF
--- a/docs/src/suitable_commands.md
+++ b/docs/src/suitable_commands.md
@@ -1,0 +1,55 @@
+# Suitable commands
+
+### help
+
+### index
+
+- [x] Manage(list, add, delete) custom plugin indexes
+
+### info
+
+- [x] Show information about a plugin via NAME & INDEX/NAME
+
+### install - Downloads, Extract & Install a plugin
+
+- [x] Install one NAME or multiple plugins [Name, Name] from default index
+- [ ] Install plugins from a newline-delimited txt file
+- [x] Install one or mutiple plugins from custom index
+- [ ] Install plugin from a local/URL custom manifest file
+- [x] If a plugin is already installed, it will be skipped
+- [x] Failiure to install a plugin will not stop the installation of other plugins
+- [ ] Cleanup Stale Installations removes the versions that aren't the current version
+
+### list
+
+- [x] show list of all charmil plugins
+- [x] show list of installed plugins
+
+### search
+
+- [ ] fuzzy search in plugins via NAME, DESCRIPTION
+
+### uninstall
+
+- [x] uninstall a plugin or list of plugins
+- [x] Failure to uninstall a plugin will result in an error and exit immediately
+- uninstall command does not support INDEX/PLUGIN syntax just specify PLUGIN
+
+### update
+
+- [x] update local copy of plugin indexes
+
+### upgrade
+
+- [x] Install a plugin or all plugins to the newer version
+
+### version
+
+- Show charmil version info & diagnostics about charmil
+- GitTagrelease name
+- GitCommit git revision ID
+- DefaultIndexURI
+- BasePath - root directory for charmil installation
+- IndexPath - path where stores the local copy of the index git repository
+- InstallPath - directory for plugin installations
+- BinPath - directory for the symbolic links to the installed plugin executables
--- a/registry/example/registry.json
+++ b/registry/example/registry.json
@@ -38,5 +38,19 @@
     "dependencies": {},
     "tags": [],
     "license": "MIT"
+  },
+  {
+    "name": "kaf",
+    "description": "Modern CLI for Apache Kafka, written in Go.",
+    "version": "v0.1.39",
+    "homepage": "https://github.com/birdayz/kaf",
+    "binaries": {
+      "darwin-amd64": "https://github.com/birdayz/kaf/releases/download/v0.1.39/kaf_0.1.39_Darwin_x86_64.tar.gz",
+      "linux-amd64": "https://github.com/birdayz/kaf/releases/download/v0.1.39/kaf_0.1.39_Linux_x86_64.tar.gz",
+      "windows": "https://github.com/birdayz/kaf/releases/download/v0.1.39/kaf_0.1.39_Windows_x86_64.tar.gz"
+    },
+    "dependencies": {},
+    "tags": [],
+    "license": "Apache License"
   }
 ]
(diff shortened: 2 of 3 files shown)


docs/src/suitable_commands.md
```diff
@@ -0,0 +1,55 @@
+# Suitable commands
+
+### help
+
+### index
+
+- [x] Manage(list, add, delete) custom plugin indexes
```
wtrocki, line 7: My take is that cli creators will hardcode their indexes. We can add that feature later.

We have two personas:

CLI creator (like ODO - who can use our SDK to bring some extensions they want to use)
CLI user (Person who uses ODO CLI)

My take is that indexes would be controlled by CLI creator and included at build time.

ankithans: You seem right, we can do this later. As other tasks look more important. As of now, we can have only one index and CLI developers can submit their plugins to that index.
later we can support their own hosted indexes

wtrocki: Not later. This practically doesn't make any sense to do. We are framework. Developers will add their indexes to code. What is the point of adding index command to their command line?

wtrocki: I knew this will sping some good conversation.

docs/src/suitable_commands.md
```diff
@@ -0,0 +1,55 @@
...
+
+### info
+
+- [x] Show information about a plugin via NAME & INDEX/NAME
```
wtrocki, line 11: Indexes would not be user facing - they will be part of the SDK IMHO
WDYT?>

docs/src/suitable_commands.md
```diff
@@ -0,0 +1,55 @@
...
+
+### install - Downloads, Extract & Install a plugin
+
+- [x] Install one NAME or multiple plugins [Name, Name] from default index
```
wtrocki, line 15: there will be no default index

docs/src/suitable_commands.md
```diff
@@ -0,0 +1,55 @@
...
+- [ ] Install plugins from a newline-delimited txt file
```
wtrocki, line 16: This is something needed. I think we need concept of profiles (they could have been hosted as kubernetes resources etc. )

Example: https://docs.openshift.com/container-platform/4.7/web_console/customizing-the-web-console.html#creating-custom-CLI-downloads_customizing-web-console

docs/src/suitable_commands.md
```diff
@@ -0,0 +1,55 @@
...
+- [x] Install one or mutiple plugins from custom index
+- [ ] Install plugin from a local/URL custom manifest file
+- [x] If a plugin is already installed, it will be skipped
+- [x] Failiure to install a plugin will not stop the installation of other plugins
+- [ ] Cleanup Stale Installations removes the versions that aren't the current version
+
+### list
+
+- [x] show list of all charmil plugins
```
wtrocki, line 25: There is no concept of charmil plugins - Charmil is framework that let others to build their plugin indexes and embed them into their CLI. We need to be clear about this difference - I think I prematurely closed #7

docs/src/suitable_commands.md
```diff
@@ -0,0 +1,55 @@
...
+
+### uninstall
+
+- [x] uninstall a plugin or list of plugins
+- [x] Failure to uninstall a plugin will result in an error and exit immediately
+- uninstall command does not support INDEX/PLUGIN syntax just specify PLUGIN
+
+### update
+
+- [x] update local copy of plugin indexes
```
wtrocki, line 40: Not needed IMHO. That would be programmatic API instead.

docs/src/suitable_commands.md
```diff
@@ -0,0 +1,55 @@
...
+
+### upgrade
+
+- [x] Install a plugin or all plugins to the newer version
+
+### version
```
wtrocki, line 46: IMHO this is like status command. Could be useful and optional to include for SDK users (cli developers)

docs/src/suitable_commands.md
ankithans: now it's all clear.

docs/src/suitable_commands.md
wtrocki: > You seem right

btw.. I'm always wrong :) question everything I say.

ankithans: yup users of CLI don't need to know indexes and don't want to add/remove indexes. Its developer's headache.

ankithans: we are not having any single index, we provide freedom to the developer to host their indexes and include them. And for this purpose hardcoding suits.

docs/src/suitable_commands.md
ankithans: users of CLI don't need to know indexes.

docs/src/suitable_commands.md
wtrocki: Yes. to make it clear - multiple indexes as feature is good - but in SDK. This PR is all about end user Commands. 
There is separate issue to cover SDK #18

docs/src/suitable_commands.md
```diff
@@ -0,0 +1,55 @@
+# Suitable commands
```
wtrocki, line 1: To merge it we can:

- Move this document to `working-group` folder 
- Adjust it to reflect all commands we want to support and flag ones that will be done in the first place.

docs/src/suitable_commands.md
ankithans: then not including this here, #18

docs/src/suitable_commands.md
ankithans: `working-group`
is it a naming convention for what we are doing or in progress?

wtrocki: It is not docs - just research doc so use whatever name to reflect it

docs/src/suitable_commands.md
ankithans: couldn't get this properly
Like in [OperatorHub](https://operatorhub.io/) Provider is the profile?
image

wtrocki: Yep :D

ankithans:
now have a look is it fine?
https://github.com/aerogear/charmil/blob/suitable-cmds/working-group/suitable_commands.md

PR #25, opened 2021-05-12
POC for creating SDK of cobra commands

ankithans:
#18
Create command in SDK and use that command in an example CLI

image

### DIFF
--- a/examples/calculator-cli/cmd/add.go
+++ b/examples/calculator-cli/cmd/add.go
@@ -1,0 +1,7 @@
+package cmd
+
+import "charmil/internal/add"
+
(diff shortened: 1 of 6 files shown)


wtrocki:
I would categorize this as learning golang and cobra rather than something that has any value in the repository. 
Let's work on top of those changes.

ankithans:
alright 👍

wtrocki:
lets integrate this with new POC branch or close

ankithans:
I will integrate #27 in a `poc` branch. I think this can be closed. It was just for learning purposes.

Issue #26, opened 2021-05-19
Charmil Roadmap


ankithans:
We need to discuss, what will be the SDK doing exactly and list of features, to proceed further.

wtrocki:
Sure. It is better to just have some base code first to leave groundwork on the discussion. 
Alternatively, we can draft some ideas here. 
I think we touchdown on how SDK will work but my take is that we simply need to define core feature. 
Krew is actually complex system if you think about it and with your level of golang it will be hard to move to that so fast. 

My suggestion is to start simple. Actually very simple and write CLI host and for the moment try to embed other commands (for example embed other CLI) - that would be possible without charmil right?

Good example would be to look at the POC that one of our colleagues wrote:
https://github.com/craicoverflow/socket

We really need an extensibility mechanism as soon as possible so starting with something that works is great.

If we ignore problem of downloading command line tool and indexing it and assume that we simply want to include modules into our binary then this might be much easier to do.

ankithans:
> My suggestion is to start simple. Actually very simple and write CLI host and for the moment try to embed other commands (for example embed other CLI) - that would be possible without charmil right?

okay got your point. Will start working on it today 👍

ankithans:
I have tried and embedding works, I created sample CLI's, installed binary and combined them (played a lot with them).
With the help of above POC.

wtrocki:
We can provide more opiniated layer for CLI by enforcing:

- Configuration
- Logging and printing messages
- Error handling and validation
- Internationalization
- Security (Authorization) - I need to think about this

namit-chandwani:
@wtrocki Very well summarized 👏
I agree with Ankit, especially on the clarity front. This appropriately depicts where we are today and how much we have to accomplish over this summer.

We'll start working now while keeping this roadmap in mind.

wtrocki:
Love your confidence!
So we have two spikes for doing core - no coincidence here ;) Pick one and create issue for deep dive discussion and implementation.

namit-chandwani:
> @ankithans have been exploring this one already so if you still want it then @namit-chandwani  can take wrapper approach.
> 
> 

I talked to @ankithans, he wants to go with the 2nd approach (wrapper). So I'll take the first one and create an issue on it as mentioned.

wtrocki:
Great job! Lets move all discussions to other issues so we will keep this one dedicated to roadmap.

[Offtopic]
[...]

PR #27, opened 2021-05-26
WIP: embeding commands poc

ankithans:
## Issues
- embeding multiple plugins in one host CLI (multiple yaml files)
- Commands are interfering each other when multiple commands are present in yaml
- we are assuming that user's system has CLI installed which is to be embedded

## steps to run
1. Run `make`
2. `./embed`

taken help from: https://github.com/craicoverflow/socket

#26 #25


LICENSE
```diff
@@ -1,202 +1,202 @@
-
```
wtrocki, line 1: I do not see any change here :D

Also license file missing owner which is wrong :D

cmd/cmd.go
```diff
@@ -1,72 +1,72 @@
...
+	"github.com/spf13/viper"
+)
+
+var cfgFile string
+
+// rootCmd represents the base command when called without any subcommands
+var rootCmd = &cobra.Command{
+	Use:   "charmil",
+	Short: "A brief description of your application",
+	Long: `A longer description that spans multiple lines and likely contains
```
wtrocki, line 20: Let's keep things refined later. Better to have empty string rather than giberish.

LICENSE
ankithans: done

cmd/cmd.go
ankithans: done

cmd/extensions/cmd.go
```diff
@@ -1,39 +1,39 @@
...
+	"github.com/aerogear/charmil/cmd/extensions/installed"
+	"github.com/aerogear/charmil/cmd/extensions/list"
+	"github.com/aerogear/charmil/cmd/extensions/remove"
+	"github.com/spf13/cobra"
+)
+
+// ExtensionsCmd represents the extensions command
+var ExtensionsCmd = &cobra.Command{
+	Use:   "extensions",
+	Short: "A brief description of your command",
```
wtrocki: :) see above ^^

ankithans: done

cmd/extensions/cmd.go
```diff
@@ -1,39 +1,39 @@
...
+Cobra is a CLI library for Go that empowers applications.
+This application is a tool to generate the needed files
+to quickly create a Cobra application.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		fmt.Println("extensions called")
+	},
+}
+
+func init() {
+	ExtensionsCmd.AddCommand(install.InstallCmd, installed.InstalledCmd, list.ListCmd, remove.RemoveCmd)
```
wtrocki, line 23: I think this is good but imagine scenario that:


Root command defines some of the types or factories (top level API) that extensions can consume. 

Those would be for logging, localizer etc. Example:
https://github.com/redhat-developer/app-services-cli/blob/main/pkg/cmd/factory/factory.go#L13-L24

So to realize that architecture we need:

- plugins-core (provides common API for host and extensions to use)
- plugins extension (provides example implementation of extension that uses structure passed from host)
- Host itself that imports extensions into go mod and have some example how plugins core structures are initialized.

cmd/extensions/installed/cmd.go
```diff
@@ -1,34 +1,34 @@
...
+Cobra is a CLI library for Go that empowers applications.
+This application is a tool to generate the needed files
+to quickly create a Cobra application.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		fmt.Println("installed called")
+	},
+}
+
+func init() {
+	// Here you will define your flags and configuration settings.
```
wtrocki: No comments please. Let's be vigilant to not push any half backed code. 
This is WIP but flagging it early.

wtrocki:
This is good start confirming that we have great understanding of basics, but we are at 10% of the POC progress.

wtrocki:
Btw.. No need to instantly address all the comments. Take it slow and easy. 
Let me know if you have questions about:

```
So to realize that architecture we need:

plugins-core (provides common API for host and extensions to use)
plugins extension (provides example implementation of extension that uses structure passed from host)
Host itself that imports extensions into go mod and have some example how plugins core structures are initialized.

ankithans:
According to my understanding:
plugin core will be like plugin_loader and plugin structure
plugin extension is embeding multiple CLI's using plugin core
> Host itself that imports extensions into go mod and have some example how plugins core structures are initialized. 

Can you explain this more?

wtrocki:
Yes. Core needs to have structure definitions that will be shared. 
What we have right now is 2 packages one depending on each other.. There is simply nothing here to make charmil bring value, but once we introduce core we can bring an opinionated layer of tools that others might use - config abstractions etc.

Look again how this factory works :)

https://github.com/redhat-developer/app-services-cli/blob/main/pkg/cmd/factory/factory.go#L13-L24

ankithans:
okay, now I got the idea about factory. Thanks 👍
working on this PR for now

ankithans:
@wtrocki can we merge this for now? I want to work on present issues on the top of it one by one.

wtrocki:
I personally do not see value of merging yet. Lets work on some feature branch for a while.

We do not want half backed code in main and POC can be done in feature branch - poc or something like this. PRs can still be fast and done against poc branch instead of main. 

There is nothing that have functional factor or significanlty shows the vision or architecture yet

wtrocki:
Little bit of background. Based on my experience in the past fast merges to main branch gived students some form of validation for the work, but we never actually had anything functional in those and ended with looooots of time wasted to cleanup things..

Main alrady has some questionable elements.
Lets work a way in small PRs but keep main clean.


We need to explore what is charmil going to offer for users and it is guaranteed that first approach might not be the right etc.

ankithans:
> fast merges to main branch gived students some form of validation for the work, but we never actually had anything functional

Yup, I can feel this now. Same was going to happen. 
We will work on ```poc``` branch till we have something valuable and refined architecture.

PR #37, opened 2021-06-02
wrapper for external CLI's 

ankithans:
- define the command structure
- create a cobra command based on that struct
```go
type CommandConfig struct {
	Name             string
	MapsTo           ArgsConfig
	Flags            []FlagConfig
	ShortDescription string
	Examples         string
}
```
- Add the above created cobra command into Host CLI

reference - https://github.com/aerogear/charmil/blob/wrapper-poc/pkg/pluginloader/loader.go
discussion #32


ankithans:
off-topic: Is it better to squash commits before merging into main?

pkg/pluginloader/loader.go
```diff
@@ -0,0 +1,84 @@
...
+
+// TODO:
+// Unmarshal the YAML
+// Extract the commands using plugin.go (command structure)
+
+// Create a Cobra Command
+func createCommand(cmdStruct *CommandConfig) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     cmdStruct.Name,
+		Short:   cmdStruct.ShortDescription,
```
wtrocki, line 21: Let's review what properties of command we want to expose later (separate issue) - there is validation etc.

pkg/pluginloader/loader.go
```diff
@@ -0,0 +1,84 @@
...
+		Example: cmdStruct.Examples,
+		Args:    cobra.ExactArgs(len(cmdStruct.MapsTo.Args)),
```
wtrocki, line 23: It is not clear for me how args mapping work here.

ankithans: at line 23
cobra.ExactArgs() takes in the number of arguments. That's what we are providing here.
and at line 25 argument mapping takes place

pkg/pluginloader/loader.go
```diff
@@ -0,0 +1,84 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			args = append([]string{cmdStruct.MapsTo.Subcommand}, args...)
+			c := exec.Command(cmdStruct.MapsTo.Name, args...)
```
wtrocki, line 26: I would preffer us to start with calling  method properly. Then we can inject different ways to exec commands that can use os.exec as well. This is called #premature-specification and affected your entire POC/Architecture.

This way arguments are not standardized, secondly we do not have ability to strongly type them on the plugin.
Let me know if you know what I mean as this is hard to explain 

TL;DR - let's not execute external process and call function
- Let's see if we can specify arguments in the way that they are clearly defined on plugin

pkg/pluginloader/loader.go
```diff
@@ -0,0 +1,84 @@
...
+		},
+	}
+
+	// add Flags
+	if cmdStruct.Flags != nil && len(cmdStruct.Flags) > 0 {
+		for _, f := range cmdStruct.Flags {
+			fs := cmd.Flags()
+			switch f.Type {
+			case "string":
+				fs.StringP(f.Name, f.Alias, f.DefaultValue, f.Description)
```
wtrocki, line 41: What would be the way for us to retrieve values. Can we use pointers

pkg/pluginloader/plugin.go
```diff
@@ -0,0 +1,28 @@
+package pluginloader
+
+type PluginConfig struct {
```
wtrocki, line 3: Love it. Documentation is good to have from beggining

pkg/pluginloader/loader.go
wtrocki: Calling external CLI is a very cool although a commonly resolved problem in the golang ecosystem.
I do not want to steal cool feature from you and we can keep this as capability, but I did not think that people will use charmil for that where you can have this:

https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/plugin/plugin.go

https://github.com/openshift-online/ocm-cli/blob/master/pkg/plugin/plugin.go

https://github.com/openshift/odo/blob/main/pkg/odo/cli/plugins/plugin_handler.go

wtrocki:
World class job! Finally we have something that we can discuss and sculpture into what we want!
Amazing work @ankithans ! I feel like you are rockstar already

pkg/pluginloader/loader.go
wtrocki: Yes. This is very specific to the execution of separate processes and it will not allow us to enforce specific argument formats etc.

ankithans:
# Can't wrap flags in with plugin
## Expected Behaviour
Flags should be mapped to the Plugin CLI

## Current Behaviour
Flags are created for the new Cobra command, but they are not mapped. They are not linked to Plugin CLI.

## Possible Solution
We might have to use `exec` package(for running external commands/processes) as we are using for mapping commands.

## Possible Implementation
1. Use flag package for parsing the command. Flag Package will split the command into an array.
2.  Output from `flag.Parse()` is `[add 1.1 2.1 --float]`
3. Check the last element of array, if that starts with `-` or `--` (then it must be a flag)
4. Append the flags to the command while using `exec` package

```go
// Execute external process/command
func Execute(cmdStruct *CommandConfig, executablePath string, cmdArgs []string) error {

	// If command has flags
	flag.Parse()
	tail := flag.Args()
	f := tail[len(tail)-1]
	if strings.HasPrefix(f, "-") || strings.HasPrefix(f, "--") { // checking last element in the tail array
		for _, fl := range cmdStruct.Flags {
			fmt.Println(fl.MapsTo)
			cmdArgs = append(cmdArgs, "--"+fl.MapsTo) // append flag to arguments
		}
	}

	// run the command
	cmd := exec.Command(executablePath, cmdArgs...)
	cmd.Stdout = os.Stdout
	cmd.Stderr = os.Stderr
	cmd.Stdin = os.Stdin
	return cmd.Run()
}
```
It's not a permanent fix. Need to figure out what else can I do!
May be there is something else at the last of command, then this fails(need to use loop to check which one is flag).

ankithans:
Thanks @wtrocki for suggesting me to write about the issue. May be it's not the correct approach, but at least this led from nowhere to somewhere!!

wtrocki:
@ankithans yes. If you hit any problems just leave comments. Proposed solution for flags doesnt make any sense.

We have some structure including flags defined at plugin with validation and value definition. This structure is then passed to core in order to create actual cobra structure including actual flags.

Then our functions need to reference those values.
How:

- Pointers/structure defined on the input structure that can be shared with core and plugin so core can use those as values

- [Bad] Naking Convention

- Plugin calls core wrapper for every arg like in cobra but custom.

Need to think about this more. If we cannot find good solution this means that wrapper option doesn't make much sense

Host command doesn't matter here - it is just getting number of commands that need to include.

wtrocki:
made update.
FYI @aerogear/charmil-core

wtrocki:
Little bit offtopic/non technical but basically at the start of every project sometimes we need to fail to actually build something good. We should try hard to do wrapper but if you see diminishing value and have concerns that should be sign to go back to drawing board, without any bad taste. Problem is that I feel that we might need to put more effort 🤣 and formalize why wrapper is `wrong approach` 

Good real life equivalent: Going to the moon was pointless, but what teams have learned when building moon infrastructure put the base to fund modern IT.
https://en.m.wikipedia.org/wiki/We_choose_to_go_to_the_Moon

ankithans:
> Pointers/structure defined on the input structure that can be shared with core and plugin so core can use those as pointers when building up cobra flags

When we say this we ensure that the plugin must be written with Go/cobra. And then it goes to 1st approach!
Let's discuss about the architecture of wrapper method and decide which way to go!!

# Plugin
Plugin can be any CLI made with any framework. Just it needs to be executable!

# Core

### Plugin/Command Input
Real command, arguments and flags of the plugin/CLI and what it should be mapped to
Eg. 
- `git clone https://github.com/aerogear/charmil --verbose` -> `host copy https://github.com/aerogear/charmil --debug`
- `calculator add 2.1 1.2 --float` -> `host plus 2.1 1.2 --fl`

### Downside of Wrapper (want to know if there is other way)
We are able to do above things using string manipulation only. I cannot see any other approach to do the above
Like here we append subcommand(for git clone, `clone` is subcommand) before the []args
`args = append([]string{cmdStruct.MapsTo.Subcommand}, args...)`
and similarly append the flags at end of `args` and execute the command.

If we want to avoid this, then ig we need to drop wrapper idea

### Creating the command
Command is created according to above input and maps it to actual plugin's binary using the above info(actual command, args, flags) using `exec` package.

### Host
Nothing as of now, just all the commands created above becomes part of host.

wtrocki:
> Plugin can be any CLI made with any framework. Just it needs to be executable!

Ok. So here we have problem - we had misunderstanding when specifying requirements. 
Requirements for plugins using wrapper is that they need to implement `our wrapper framework`.
I think you have brought this requirement as bonus to be able to call existing CLI's - that is not our target. 
Let's scrap this completely for now (we can do that by having `executable wrapper plugin` that understands some executables that are already installed and can call them. But this is not what we do.

With that in mind you can edit your reply above as it doesn't have sense for the approach I have mentioned

ankithans:
okay, let's scrap this. Now I know what is required. 
Working on it and will update the architecture

wtrocki:
Let's keep that separately.

wtrocki:
Closing as we no longer pursue this approach

PR #44, opened 2021-06-07
feat: add charmil-sdk poc (root validation approach)

namit-chandwani:
Based on discussion: https://github.com/aerogear/charmil/discussions/43#discussion-3399502

## Working of the current version of POC

- Currently, this POC uses the ability of Go packages that enable them to be compiled as shared object (.so) libraries and to be loaded dynamically at run time.
- The `charmil install` command (for the time being) takes the name of the plugin as argument and generates `.so` file for that plugin using the command:

    ```golang
    go build -buildmode=plugin
    ```
-  The Golang standard [plugins package](https://golang.org/pkg/plugin/) is then used to import cobra structures of the plugins into the host CLI using these `.so` files.

- This is just a temporary approach and has a few downsides to it (as discussed below). This will later be replaced by an alternate approach that suits our use case even better.

## File Structure of the POC

```code
📦Charmil-SDK-POC
 ┣ 📂cmd
 ┃ ┣ 📂charmil              (The CLI which the plugin developers will use to manage plugins for their own CLI)
 ┃ ┃ ┗ 📜main.go
 ┃ ┗ 📂hostCLI              (Example of a host CLI made using the Charmil SDK. This is included here just for ease and will be removed later)
 ┃ ┃ ┗ 📜main.go
 ┣ 📂docs                   (Documentation)
 ┣ 📂pkg
 ┃ ┣ 📂cmd
 ┃ ┃ ┗ 📂charmil            (Contains only the logic for `charmil install` command at this moment. More commands can be added later)
 ┃ ┃ ┃ ┗ 📜install.go       (A command which will install the specified plugins into developer's environment. Right now the temporary logic there is to generate the shared object of the plugin name specified, which will be later replaced by logic which will download, validate, extract and then install plugin binaries)
 ┃ ┣ 📂core                 (Empty now. Will add helper functions for building host CLI in the next version)
 ┃ ┗ 📂plugin               (Empty now. Will add plugin handler logic in the next version)
 ┣ 📂plugins                (Contains all the plugins to be used by the developer. Included in this repo just for ease, will be removed later)
 ┃ ┣ 📂pluginA              (Example plugin A which is a CLI built using cobra)
 ┃ ┃ ┗ 📜main.go
 ┃ ┣ 📂pluginB              (Example plugin B which is a CLI built using cobra)
 ┃ ┃ ┗ 📜main.go
 ┃ ┣ 📂pluginC              (Example plugin C which is a CLI built using cobra)
 ┃ ┃ ┗ 📜main.go
 ┃ ┗ 📂shared_objects       (Contains the .so files generated by the `charmil install` command. These are currently used along with the Golang standard plugins package to import plugin cobra structures into the host CLI)
 ┃ ┃ ┣ 📜pluginA.so
 ┃ ┃ ┣ 📜pluginB.so
 ┃ ┃ ┗ 📜pluginC.so
 ┣ 📜.gitignore
 ┣ 📜LICENSE
 ┣ 📜Makefile
 ┣ 📜README.md
 ┣ 📜go.mod
 ┗ 📜go.sum
```
&nbsp;
## Why we should avoid using the standard [plugins package](https://golang.org/pkg/plugin/) for our use case?

- The environment for building a Go plugin such as OS, Go language version, and dependency versions must match exactly with the host CLI.
- As of now, unloading plugins are not allowed without restarting the program.
- We cannot replace a plugin with a newer version during runtime.
- Using it, we can only build plugins on Linux, FreeBSD, and MacOS as of now.

## Commands that can be added to Charmil CLI in the future

- install
- installed
- list
- uninstall
- load
- unload
- index
  - add
  - remove
  - update

## Things yet to do

- [ ] Add plugin handler code
- [ ] Define appropriate structures to store commands along with their config for validation
- [ ] Add root validation logic


wtrocki:
not sure if loading dynamic binaries is good approach.
As you have commented it has some issues and it was originally created for different reasons.

We are looking at simplicity of distribution and golang dynamic plugins are not giving them to us.
Additionally plugins will loose ability to build cli for plugin on it's own. 

My take is that this approach is simply putting very strong restrictions to architecture. It would have been better to actually discuss that before starting work.

I would simply include plugins into code.

wtrocki:
@ankithans your take?

namit-chandwani:
> Not sure if loading dynamic binaries is good approach.
> As you have commented it has some issues and it was originally created for different reasons.
> 
> We are looking at simplicity of distribution and golang dynamic plugins are not giving them to us.
> Additionally plugins will loose ability to build cli for plugin on it's own.
> 
> My take is that this approach is putting very strong restrictions to architecture. It would have been better to actually discuss that before starting work.
> 
> I would include plugins into code.

Yes, agreed!
My original plan was just to utilize this approach for setting things initially and then switch to something different.

But you're absolutely right, sticking to a plan right from the beginning is much better than modifying it halfway through. It's ideal to first discuss the approach we'd want to take and then execute it after we're confident that it'll work for our use case.

So now I'll start exploring the alternative options that we can take and leave a comment here with the details once I'm done.

ankithans:
so after reading the disadvantages of `plugins package` in this use-case we should definitely go for some alternate option!

wtrocki:
I would look for code that drafts how interaction will work the we can sit down and think what is best

namit-chandwani:
> I would include plugins into code.

@wtrocki According to what I've understood, by this line you meant that we should consider following the approach marked as `2.3` in the discussion: https://github.com/aerogear/charmil/discussions/47#discussion-3407566, right?

wtrocki:
Yes. Closing this PR. Let's start fresh with 2.3 :)

PR #45, opened 2021-06-09
charmil as wrapper for cobra

ankithans:
## Plugin 
`plugins/`
Uses Plugin structure to create a command

## Core 
`pkg/pluginloader/`
Takes the input from plugin and create a cobra command out of it.

## Host 
`cmd/host/`
Adds above created commands as host commands

## Example: 
- create an add command in the plugin using charmil-core
```go
var AddCommand = pluginloader.CreateCommand(&pluginloader.CommandConfig{
	Name: "add",
	Args: []string{"num1", "num2"},
	Flags: []pluginloader.FlagConfig{{
		Name:         "float",
		Alias:        "f",
		DefaultValue: "false",
		Type:         "bool",
		Description:  "Add floating numbers",
	}},
	ShortDescription: "add numbers",
	Examples:         "$ host plus 1 2",
}, func(cmd1 *cobra.Command, args []string) {
	ExecuteAddition(cmd1, args)
})
```
- Add this command to host CLI

According to https://github.com/aerogear/charmil/pull/37#issuecomment-856591977


wtrocki:
@aerogear/charmil-core

cmd/host/main.go
```diff
@@ -0,0 +1,24 @@
...
+
+	"github.com/aerogear/charmil/plugins/calculator"
+	"github.com/aerogear/charmil/plugins/hi"
+	"github.com/spf13/cobra"
+)
+
+func main() {
+	cmd := &cobra.Command{
+		Use:          "Host",
+		Short:        "Host CLI for embedding commands",
```
wtrocki, line 14: Example Host

plugins/calculator/main.go
```diff
@@ -0,0 +1,62 @@
...
+	"strconv"
+
+	"github.com/aerogear/charmil/pkg/pluginloader"
+
+	"github.com/spf13/cobra"
+)
+
+var AddCommand = pluginloader.CreateCommand(&pluginloader.CommandConfig{
+	Name: "add",
+	Args: []string{"num1", "num2"},
```
wtrocki, line 14: What approach you use for arguments

wtrocki:
@ankithans please make required changes to the structure.

@aerogear/charmil-core @namit-chandwani let's review this PR so we can merge this as initial base for project

cmd/host/main.go
namit-chandwani: > @namit-chandwani maybe we can agree on structure and kinda push that to the main so you both base on the same structure and we do not have issues?

Yeah sure

plugins/calculator/main.go
ankithans: cobra accepts the arguments in array of strings `args []string`. So I kept it that way. May be we can add NumArgs option for the user.
Any other thoughts?

There might me some commands which sometimes require 1 arg sometimes 2 args. Then specifying exact number of args will be wrong!

ankithans:
pushed the required changes.

plugins/calculator/main.go
wtrocki: my take is that wrapper will not work ad it will require significant refactor to integrate existing commands. Some of those will not work with wrapper assumptions which means that we need to break API etc. instead of pushing towards something that is broken we can join forces and do proper cobra based plugin

main.go
```diff
@@ -1,7 +1,7 @@
 package main
 
-import "github.com/aerogear/charmil/cmd"
+import "fmt"
 
 func main() {
-	cmd.Execute()
+	fmt.Println("Welcome to Charmil")
```
wtrocki, line 6:
```suggestion
```

pkg/pluginloader/create.go
```diff
@@ -0,0 +1,40 @@
+package pluginloader
+
+import (
+	"strconv"
+
+	"github.com/spf13/cobra"
+)
+
+// create a new cobra command
+func CreateCommand(cmdStruct *CommandConfig, execute func(cmd *cobra.Command, args []string)) *cobra.Command {
```
wtrocki, line 10: Leftover from wrapper! I guess it is ok to leave it.

pkg/pluginloader/plugin.go
```diff
@@ -0,0 +1,17 @@
+package pluginloader
+
+type CommandConfig struct {
```
wtrocki, line 3: Leftover from wrapper? Ok to leave it for now

wtrocki (approved):

.gitignore
```diff
@@ -13,3 +13,5 @@
 
 # Dependency directories (remove the comment below to include it)
 # vendor/
+
+/host
```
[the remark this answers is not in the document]
ankithans: makes sure that binary file is not pushed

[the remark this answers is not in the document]
wtrocki: proper format will be `host`, but this surprisingly works fine :D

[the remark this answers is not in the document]
ankithans: host took the `host` folder as well 🥲 so had to do this

PR #48, opened 2021-06-14
feat: logger for output & errors

ankithans:
## Description
- implement top level factory for commands
- logger for printing output and error
   - not a complete logger though, just implemented two methods
   - lot of useful methods & options can be added in logger
- use factory with logger in plugins
- import plugins into Host

want to validate, if this is a fine approach?


ankithans:
@wtrocki

wtrocki (approved):

wtrocki:
Good start

pkg/logging/logger.go
```diff
@@ -0,0 +1,6 @@
+package logging
+
+type Logger interface {
```
wtrocki, line 3: Docs please

pkg/logging/std_logger.go
```diff
@@ -0,0 +1,53 @@
...
+		logger.outStream = os.Stdout
+	}
+	if logger.errStream == nil {
+		logger.errStream = os.Stderr
+	}
+
+	return
+}
+
+func (l *StdLogger) Output(args ...interface{}) {
```
wtrocki, line 47: Docs

PR #49, opened 2021-06-15
feat: add updated poc

namit-chandwani:
Took help from: https://github.com/redhat-developer/app-services-cli


namit-chandwani:
@wtrocki

wtrocki:
This is duplicated and replicated with what we have in main. Let's rebase on top of master. 
If you looking to work on specific feature let's do issue first to avoid clashes with others.

CC @aerogear/charmil-core

PR #52, opened 2021-06-17
Add i18n support & improve logging

### LINKED ISSUE #50: Add internationalization support(i18n) in commands
## Feature Description
Support multiple languages for the CLI commands, subcommands, flags, logging etc. 

## Proposed Solution
- create package for localization in `charmil-core` and add it to factory
- plugin developer use factory to localize everything and provide locals for the languages to be supported

## Additional Context
good examples to refer:
https://github.com/redhat-developer/app-services-cli/tree/main/pkg/localize
https://github.com/apache/openwhisk-wskdeploy/tree/master/wski18n

### LINKED ISSUE #51: Add methods for unified logging in commands
## Feature Description
Following methods can be added to ensure the logging and printing messages in unified way
Debug, Info, Warn, Error, Fatal, Panic, Trace, etc. using `fmt.Fprint()`, `fmt.Fprintf()`, `fmt.Fprintln()` functions

ankithans:
i18n support from factory, provides each plugin to implement localization
- Plugin once initializes the localization package providing the config details
- can localize text items by id defined in locals file
- improve logging methods
- provide API docs

closes #50 
closes #51


wtrocki:
I would like us to write documentation first or at leas always add documentation to PR.

examples/plugins/calculator/main.go
```diff
@@ -4,22 +4,34 @@ import (
...
+	Localize localize.Localizer
 }
 
-func RootCommand(f *factory.Factory) *cobra.Command {
+func RootCommand() *cobra.Command {
+
+	loc, err := localize.InitLocalizer(localize.Config{Language: language.English, Path: "examples/plugins/calculator/locals/en/en.yaml", Format: "yaml"})
+
+	if err != nil {
+		panic(err)
```
wtrocki, line 23: Panic is not going to provide good output. Never use panic.
Especially that logger is there ( However undocumented )

ankithans:
> I would like us to write documentation first or at least always add documentation to PR.

Sure, I have written some comments in pkg. 
Or you are saying about md docs about feats?

examples/plugins/calculator/main.go
ankithans: Okay 👍
Have to add some more methods in logger. Will document it.

wtrocki:
Yes.. There are two types of docs:

- In code for in IDE exploration to read about what code they write does
- For developers to actually write that code in IDE in first place

We need.. both :D

wtrocki:
@ankithans take look into discussions - some of the topis there were directly inspired by your PR.

examples/plugins/calculator/main.go
wtrocki: Example should be smaller. 

If we basically copy paste other library example and implement it what is the value we provide?

wtrocki:
I looked at it and it feels that this PR is ok but also it is not giving us anything:

We have an undocumented implementation of the another library that is scatered across different examples that developers will most likely struggle to see and adopt. I will need to think about it how to deliver this to make sense.

If our approach will focus on simply dropping some libraries into factory - what is the difference with someone using factory pattern in their CLI?

ankithans:
Also, we should make the features simple to use(focus on UX of lib as you mentioned before) to make sure that we are not making it harder to implement for the developers.

Documentation and smaller examples would show users how to use the library's functionality. We can take one example command which will utilize all the functions of cobra like flags, args, etc. and implement the library into that while documenting.

Let me work on docs & examples of logger & i18n.

examples/plugins/calculator/main.go
ankithans: will not say it exact copy paste, but yeah it is. When you see a good approach in the front of you, it blocks the way for other things. But I will try to minimize these things in further contributions!

ankithans:
@wtrocki I have provided the docs to the API. And to the example plugin as well
Should I do md docs for logger & i18n in this PR?

ankithans:
@wtrocki ready for review. 
Provided docs.

docs/src/charmil_factory.md
```diff
@@ -0,0 +1,36 @@
+## Charmil Factory
+The Charmil Factory offers you one-stop access to all of the Charmil packages all at once. So you just need to initialize factory in your cobra command and start using logging, i18n, etc. 
+```go
+import github.com/aerogear/charmil/pkg/factory
+```
+
+## How to use
+
+1. Create a clean cobra command for your plugin
```
wtrocki, line 9:
```suggestion
1. Create a cobra command for your plugin

```
we do not like dirty cobras? racial discrimination

docs/src/charmil_factory.md
```diff
@@ -0,0 +1,36 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return nil
+		},
+	}
+
+	return cmd, nil
+}
+```
+
+2. Create a default factory instance in your command, which requires the localizer to be passed
```
wtrocki, line 24:
```suggestion
2. Create a factory instance in your command, which requires the localizer to be passed

```

docs/src/charmil_factory.md
ankithans: done

docs/src/charmil_factory.md
ankithans: done

ankithans:
@wtrocki addressed comments

wtrocki:
Sure. Since this is giving us some base let's merge but I would encourage other mentolrs to still comment on it so we can create separate PR.

Issue #58, opened 2021-06-17
Validator to ensure standard for commands


wtrocki:
So we removed the wrapper approach as a feasible way to drive our architecture, but despite not having it 
we still want to control some of the aspects of the commands. 

My take is that for this purpose we could have a linter/validator solution where we can specify different rules that will be interacting thru cobra commands and see if they are valid etc. We will provide default recommendations/ruleset but there should be also the possibility to provide a validator in the near future. 

This is the top-level issue we can create task issues once we discuss more how this would look like.

@ankithans some bits of your wrapper implementation might be very useful for that.

wtrocki:
We need resarch first:

1) If something like this is not actually available right now 
2) if we want to write validator ourselves or use existing linting solution like: https://github.com/golangci/golangci-lint - not sure if that would even allow us to do what we want.

3) Discuss examples of rules we want to apply:

Example: Each command should have long and short description and example.

wtrocki:
@ankithans Check comment above. I would love to get your replies to it :)

> We need research first

ankithans:
> Discuss examples of rules we want to apply

as you already mentioned some of the rules, copying them here
- Command has examples 
- Short and Long description
- Each argument has a description
- Command has long and short description defined
- Command uses single positional argument (no multiple positional arguments) etc.

more
- Command has a Run function
- Defined number of arguments in command

For now couldn't think more useful

We can use this for docs/help (examples and Long)
https://github.com/MakeNowJust/heredoc

---
> if we want to write a validator ourselves or use an existing linting solution like golangci-lint

To use golangci-lint, don't we have to write validators ourselves using AST for cobra specific structure and then integrating that with golangci-lint?

---
one more ques:
if there is a possibility that we could extend the functions from cobra library not wrapping?

wtrocki:
> one more ques:
if there is a possibility that we could extend the functions from cobra library not wrapping

What you mean by this. More detailed question

ankithans:
In dart there is a feature to extend the library - modify, overriding existing methods

wtrocki:
Sill no clue.. 
Done quick research and we can do:

1) Ast processor (go has core lib that enables and most of linters use it). It will work at compile time etc.

Benefits: It will be easy to make this popular, blog post it etc.

Drawbacks: Overkill for our needs, harder to add new rules 

2) In house cobra structure analiser with rules.
It can be cli on its own that needs to be compiled and will process rules and validate stuff.
[...]

PR #63, opened 2021-06-17
feat: add alternative approach for base SDK

namit-chandwani:
WIP. Will add PR description here shortly


wtrocki:
> `alternate approach` 

Stack overflow. Maximum number of approaches reached.

Makefile
```diff
@@ -1,5 +1,14 @@
-run:
-	go run ./examples/host
+.DEFAULT_GOAL := build
+CHARMIL_BIN_FILE = charmil
```
wtrocki, line 2: charmil CLI does what?

cmd/charmil/main.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/spf13/cobra"
+)
+
+func main() {
+	cmd := &cobra.Command{
+		Use:          "charmil",
+		Short:        "Commands for managing plugins",
+		SilenceUsage: true,
+		Run: func(cmd *cobra.Command, args []string) {
+			fmt.Println("Welcome to Charmil")
```
wtrocki, line 16: Screenshot 2021-06-17 at 16 19 22

examples/plugins/calculator/main.go
```diff
@@ -1,73 +0,0 @@
-package calculator
```
wtrocki, line 1: Try to avoid deleting some one else code in POC's to not cause big changesets. Removing and cleanup can happen later.

pkg/core/plugin.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"fmt"
+	"log"
+
+	"github.com/spf13/cobra"
+)
+
+var plugins = make(map[string]Plugin)
+
+func Register(name string, plugin Plugin) {
```
wtrocki, line 12: So basically instead of using go mod to inexplicitly 



This is clever, but I want to know what we getting from this? Is this art for the art or there is any benefit coming from
registration of plugins - which is extra hassle for something that can be done using golang and cobra (importing root command and adding it to the list in any place we want.

pkg/core/plugin.go
```diff
@@ -0,0 +1,30 @@
...
+	if plugin == nil {
+		log.Fatal("Undefined plugin")
+	}
+	if _, isDup := plugins[name]; isDup {
+		log.Fatal("Plugin with this name already exists")
+	}
+	plugins[name] = plugin
+}
+
+func GetRootCmd(pluginName string) (*cobra.Command, error) {
```
wtrocki, line 22: I'm adversary for and developer-facing API that bases on unstructured content - Plugin name as string has no way to get validation on compile time. This is huge bummer because without this we can bring plugin root command and get it validated against cobra at compile time. 

Again this should have been discussion - but code makes it easier to reply so it is good as well

Makefile
namit-chandwani: This is outdated code. 
Remember we were once planning to include a Charmil CLI which SDK users can interact with? This was meant for that.

But I've removed it now, since there have been changes in our plans and we're no longer going to implement this

cmd/charmil/main.go
```diff
@@ -0,0 +1,23 @@
...
+	"fmt"
+	"log"
+
+	"github.com/spf13/cobra"
+)
+
+func main() {
+	cmd := &cobra.Command{
+		Use:          "charmil",
+		Short:        "Commands for managing plugins",
```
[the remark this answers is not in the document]
namit-chandwani: Outdated code (as mentioned in comment https://github.com/aerogear/charmil/pull/63#discussion_r653680593).

Have removed it now

cmd/charmil/main.go
namit-chandwani: https://github.com/aerogear/charmil/pull/63#discussion_r653681452

Okay, will use logger everywhere.

examples/plugins/calculator/main.go
namit-chandwani: Coincidently, even I had developed a calculator CLI earlier which I'm using as an example plugin now. So did that to avoid any confusion.

Will add it back

namit-chandwani:
> > `alternate approach`
> 
> Stack overflow. Maximum number of approaches reached.
> 
> Edit:
> 
> You probably meant alternative? Alternate means recurrent

Yeah I know, don't know why I wrote that. Fixed it now 😅

pkg/core/types.go
```diff
@@ -0,0 +1,7 @@
+package core
+
+import "github.com/spf13/cobra"
+
+type Plugin interface {
+	CreateRootCmd( /*TODO: Define config struct to add as arg here*/ ) *cobra.Command
```
wtrocki, line 6: Benefit of wrapping cobra root CLI?

I would say we could do validate that but my preference is to do static linting rather than validating thing at runtime.
See #58 

It is hard to wrap my head around this approach without seeing config struct and since I do not see it here this probably means that you are still figuring it out.

wtrocki: Another bummer here is that by forcing us to have `ROOT` we instantly telling people that we cannot have 2 root commands or set of the commands that people can bring into their CLI. etc.

pkg/core/plugin.go
namit-chandwani: Now that I think, yes, registering plugins isn't adding any extra value to our SDK

wtrocki: As mentioned in discussions we can always go back to this spikes. Sometimes we need to let go and move on. Lets close this PR, create issues we spoke and push forward with whatever will be the most challenging for you.

wtrocki: Not having any ideas around 3am is because it is 3am and not because this approach is bad. Get some sleep and come back to this later

namit-chandwani: Yes, you're absolutely right.

Will close this PR for now and shift my focus to the current issues/discussions for the time being

PR #80, opened 2021-06-22
docs: suggested changes to i18n and logger docs

jjkiely:
@ankithans your PR https://github.com/aerogear/charmil/pull/52 was merged before I got a chance to review it, I think you've done some great work but I had a few suggestions regarding the docs (mostly around spelling) so I've pushed them here, let me know if you disagree with anything :smile:

@aerogear/charmil-core


docs/src/charmil_logger.md
```diff
@@ -13,7 +13,7 @@ Errorln(args ...interface{})
 ```
 
 ## How to use
-When the factory is initialised, the logger is provided by default. To know about factory visit [here](charmil_factory.md)
+When the factory is initialized, the logger is provided by default. To know about factory visit [here](charmil_factory.md)
```
ankithans, line 16: awesome @jjkiely 
there were a lot of mistakes 😅 Thanks for correcting them!

ankithans (approved):

docs/src/charmil_logger.md
jjkiely: No problem 😊
I had to use a dictionary to figure out of locale or local was correct in this context 😂

PR #82, opened 2021-06-24
feat: validator for cobra cmds

### LINKED ISSUE #58: Validator to ensure standard for commands

ankithans:
## Methods
Validates the command and it's descendants
- `func Validate(cmd *cobra.Command) error` // handles default errors declared in validator package
- `func ValidateCustom(cmd *cobra.Command, handleErrors func() error) error` // on the top of default errors, user can define custom errors

## To decide
- what aspects we want to control
  - For now : Use, Short, Long, Example, Args
- How do we like this validator package to be served
  - from factory?
  - seperate from sdk?

@aerogear/charmil-core 
closes #58


wtrocki:
```
How do we like this validator package to be served
from factory?
seperate from sdk?
``` 

I think I have provided info for that in issues and chat. Basically the best would be for this to be completely separate library. 
And as mentioned this path doesn't have very clear way to run this without investigation.
My take is that could be helper golang unit tests library and rules should be actual unit tests, but please research this properly.

wtrocki:
I would really love us to come with rules that can declare the presence of the field. Length of the field etc.
This way we can build some generic tool that cover cobra structures and came up with more advanced rules. 

I would love to see this done this week so we can integrate that with the RHOAS CLI next week as unit test or whatever you figure out.

examples/plugins/date/main.go
```diff
@@ -49,5 +53,11 @@ func DateCommand() (*cobra.Command, error) {
 		},
 	}
 
+	// default validation provided by validator package
+	validationErr := validator.Validate(cmd)
```
wtrocki, line 57: Yes.. while this is ok. It makes no sense. 

Validation is development feature but we are using it at runtime. This needs to change.

validator/std_validator.go
```diff
@@ -0,0 +1,70 @@
...
+		err := handleDefaultErrors(child)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// ValidateCustom will allow user to
```
wtrocki, line 29: This is wrong. Every rule should be custom rule. Create interface for rule with single method that validates single command and then traverse thru nested structure.  This is easier than current architecture and will allow anyone to drop any rules - users will always need to define rules somehow or override them, but it should not require to actually write them - for example fieldExist or string length rule needs to have some arguments to be defined.

validator/std_validator.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	return nil
+}
+
+// handleDefaultErrors handles the default errors
+func handleDefaultErrors(cmd *cobra.Command) error {
+	cmdPath := cmd.CommandPath()
+
+	if len(cmd.Use) < 1 {
```
wtrocki, line 49: Makes no sense in number of dimensions:


- we are missing exact location that can be used to pinpoint file that causing issues
- we are returning only single issue (rather than numerous issues during linting)
- Hardcoded values.

wtrocki:
please rebase conflicts. I have done some investigations and I feel like unit testing framework with assertions could work, but number of rules we would be able to pull will be limited to structure checks.

I think that is fine - tried to do AST linter and while you can do the same - it is way harder to write rules

ankithans:
@wtrocki conflicts resolved 
so should I proceed with AST linter? Hit and trial is good approach for doing AST?

ankithans:
how will the user add the custom rules. Will he be writing rules with AST? 
or the user will be only using the rules provided by us, selecting what they want to apply?

wtrocki:
No.. Lets finalize this PR properly. We need to lower expectations to adjust to progress. AST linter is 3-4 times more work

wtrocki:
> how will the user add the custom rules. Will he be writing rules with AST?

 create golang interface representing rule and build number of rules based on them. Users could add their own

wtrocki:
@ankithans I think it will be good to have class and sequence diagram (can be on piece of paper) to visualize what we want to do.

validator/length.go
```diff
@@ -0,0 +1,97 @@
...
+	return errors
+}
+
+func validateCobraCommand(cmd *cobra.Command, l *LengthRule) []error {
+	var errors []error
+
+	cmdPath := cmd.CommandPath()
+
+	use := cmd.Use
+	useErr := validateField(l.Use, use, cmdPath)
```
wtrocki, line 41: This part might bring some maintenance challenges and breaks DRY rule.

wtrocki: We can use reflect package to get values by strings there is no need to repeat them

validator/length.go
```diff
@@ -0,0 +1,97 @@
+package validator
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+type LengthRule struct {
+	Use     Limit
```
wtrocki, line 10: While this is clever I think it restricts us how rules are build. 
My take is that rule:

- Has name
- Has Function that does something - Doesn't have any fields 
- Has specific arguments that configure how rule works.

In this case name is defined as Struct but I think we also need name as constant.

this way we can build `LengthRule` with dictionary of key values, where key could be cobra structure key like `Use` and values might be functions, min, max etc.
```
LengthRule(Arguments{
"Use": Limit{3,10},
"Short": Limit{3,10},
})
```

examples/plugins/date/main.go
```diff
@@ -53,11 +54,14 @@ func DateCommand() (*cobra.Command, error) {
 		},
 	}
 
-	// default validation provided by validator package
-	validationErr := validator.Validate(cmd)
-	if validationErr != nil {
-		return nil, validationErr
+	var r validator.Rule = &validator.LengthRule{
```
wtrocki, line 57: This approach as number of drawbacks:

- it is fixed to the point where we need to redeclare every field 
- cannot be declarative
- Doesn't provide us base for adding other rules (missing interface for rule)

wtrocki:
Amazing work! This is getting closer - but still needs to be refined and we need more rules. Nice to see some non trivial code happening

validator/length.go
ankithans:
```go
type LengthRuleMap map[string]Limit // instead of struct
```
```go	
// user
var r1 validator.Rule = &validator.LengthRuleMap{
	"Use":     validator.Limit{Min: 1, Max: 5},
	"Short":   validator.Limit{Min: 4, Max: 5},
	"Long":    validator.Limit{Min: 5, Max: 20},
	"Example": validator.Limit{Min: 5, Max: 30},
}
```
yes this helps in traversing and don't break DRY
here we will need to validate the `key` in dict, to map with cobra command struct

wtrocki:
Rough example showing what I meant by using reflection:
https://play.golang.org/p/Y9tSo1k8FsS

ankithans:
@aerogear/charmil-core 
we have length & mustPresent validator for now. Any more ideas? what would we like to validate in cobra command?

wtrocki:
lets start with those two for now

validator/length.go
```diff
@@ -0,0 +1,77 @@
...
+	Min, Max int
+}
+
+func (l *LengthRule) Validate(cmd *cobra.Command) []error {
+
+	var errors []error
+	err := validateLength(cmd, l)
+	errors = append(errors, err...)
+
+	for _, child := range cmd.Commands() {
```
wtrocki, line 24: we are doing only one level of commands. we should check all commands including children of children

validator/length.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	return errors
+}
+
+func validateLength(cmd *cobra.Command, l *LengthRule) []error {
+	var errors []error
+
+	for fieldName, limits := range l.Limits {
+		reflectValue := reflect.ValueOf(cmd).Elem().FieldByName(fieldName).String()
```
wtrocki, line 37: This is going to blow up..

wtrocki, line 37: no error handling/input validation

validator/length.go
```diff
@@ -0,0 +1,77 @@
...
+func validateField(limit Limit, value string, path string) error {
+	length := len(value)
+
+	_, err := isLimitSet(limit)
+	if err != nil {
+		return err
+	}
+
+	if length < limit.Min {
+		return fmt.Errorf("%s in %s: length should be atleast %d", value, path, limit.Min)
```
wtrocki, line 56: typo atleast

validator/length.go
```diff
@@ -0,0 +1,77 @@
...
+
+	return nil
+}
+
+func isLimitSet(limit Limit) (bool, error) {
+	if limit.Max < 0 || limit.Min < 0 {
+		return true, fmt.Errorf("max and min must be greater than 0")
+	}
+	if limit.Max == 0 && limit.Min == 0 {
+		return false, fmt.Errorf("limit not set")
```
wtrocki, line 70: lack of good message patterns.

Limit to what is not set? Where?

Would use rule name for all rule specific errors.

validator/length.go
```diff
@@ -0,0 +1,77 @@
...
+		err := validateField(limits, reflectValue, cmd.CommandPath())
+		if err != nil {
+			errors = append(errors, err)
+		}
+	}
+	return errors
+}
+
+func validateField(limit Limit, value string, path string) error {
+	length := len(value)
```
wtrocki, line 48: Lets add verbose mode that prints debug info for each rule cmd field

validator/must_present.go
```diff
@@ -0,0 +1,51 @@
...
+
+type MustPresentRule struct {
+	Fields []string
+}
+
+func (p *MustPresentRule) Validate(cmd *cobra.Command) []error {
+
+	var errors []error
+	err := validateMustPresent(cmd, p)
+	errors = append(errors, err...)
```
wtrocki, line 18: for visiting graph/tree best to pass info object and aggregate stats on it.
Number of commands checked. Errors.
Currently errors are string - we need this as struct with much more info

validator/must_present.go
```diff
@@ -0,0 +1,51 @@
...
+	var errors []error
+
+	for _, field := range p.Fields {
+		reflectValue := reflect.ValueOf(cmd).Elem().FieldByName(field)
+		errors = append(errors, validateByType(&reflectValue, field, cmd.CommandPath())...)
+	}
+	return errors
+}
+
+func validateByType(reflectValue *reflect.Value, field string, path string) []error {
```
wtrocki, line 38: What this does? Not sure why we need it

ankithans: cobra.Command struct has different types of attributes. Hence to address all them. I did this. 
Like something is string, bool, []string, slice, func 
eg - Args is func, SilenceUsage is bool

by writing this function we can handle `mustPresent` for all the types present in cobra.Command

wtrocki:
Talked with stackholders.
We need:

- useMatches that accept valid regexp that checks conmand name
- Arguments validators - provide global arguments that should not be redefined

validator/must_present.go
wtrocki: ok

ankithans:
> Arguments validators - provide global arguments that should not be redefined

can you give a small example to understand?

wtrocki:
Cobra can have number of issues with arguments - arguments defined on child can shadow global ones that were defined in the root. Arguments can override shorthands etc.

ankithans:
okay clear!
need to research how to proceed with this one

ankithans:
@wtrocki @aerogear/charmil-core 
addressed above comments. ready for review

ankithans:
dev docs added

examples/plugins/date/main.go
```diff
@@ -49,5 +54,23 @@ func DateCommand() (*cobra.Command, error) {
...
+	// 		"Short": {Min: 4, Max: 10},
+	// 		"Long":  {Min: 5, Max: 20},
+	// 		"Not":   {Min: 2, Max: 22},
+	// 	},
+	// }
+
+	// errr := r1.Validate(cmd, true)
+	// fmt.Println(errr)
+
+	var r2 validator.Rule = &validator.MustPresent{
```
wtrocki, line 69: Why we run validator as part of the command?

validator/error.go
```diff
@@ -0,0 +1,23 @@
+package validator
+
+import "errors"
+
+// Define errors
+var (
```
wtrocki, line 6: [Style][ Errors can be local to the rules, for maintenance and redability.
This is only my own opinion

examples/plugins/date/main.go
ankithans: I did this just to test it's working

validator/length.go
```diff
@@ -0,0 +1,127 @@
...
+	Limits map[string]Limit
+}
+
+// Limit defines min, max length of string
+type Limit struct {
+	Min, Max int
+}
+
+// Validate implements the Rule interface
+func (l *Length) Validate(cmd *cobra.Command, verbose bool) []Error {
```
wtrocki, line 23: We should use interface for validator

wtrocki: Verbose should be an config structure.

validator/error.go
ankithans: So let's not define them and do hardcode?

validator/length.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// ValidateHelper returns errors of type []Error
+// if length is not validated
+func (l *Length) ValidateHelper(cmd *cobra.Command, verbose bool, info stats) []Error {
+
+	// validate the root command
+	err := validateLength(cmd, l, verbose)
+	// record stats
+	info.num++
```
wtrocki, line 37: totalTested?

validator/length.go
```diff
@@ -0,0 +1,127 @@
...
+	info.num_failed += len(err)
```
wtrocki, line 38: validationErrors

validator/length.go
```diff
@@ -0,0 +1,127 @@
...
+	info.errors = append(info.errors, err...)
```
wtrocki, line 39: We need errors to have more than string - struct with cobra command and rule name that caused it

ankithans: yes i did it
```go
type Error struct {
	Name string
	Err  error
	Rule string
}
```

validator/length.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// traverse descendents of cmd
+	for _, child := range cmd.Commands() {
+
+		// base case
+		if !child.IsAvailableCommand() || child.IsAdditionalHelpTopicCommand() {
```
wtrocki, line 45: Add issue to provide config with array of commands to skip/ignore

validator/length.go
```diff
@@ -0,0 +1,127 @@
...
+// provided and compare with the limit
+func validateLength(cmd *cobra.Command, l *Length, verbose bool) []Error {
+	var errors []Error
+
+	for fieldName, limits := range l.Limits {
+		// reflects the fieldName in cobra.Command struct
+		reflectValue := reflect.ValueOf(cmd).Elem().FieldByName(fieldName)
+
+		// if the defined fieldName doesn't exist in cobra.Command
+		if reflectValue.String() == "<invalid Value>" {
```
wtrocki, line 73: I think that reflect has method for checking this before we call String

validator/length.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// check if valid limit is set
+	_, err := isLimitSet(limit)
+	if err.Err != nil {
+		return err
+	}
+
+	// prints additional info in debug mode
+	if verbose {
+		fmt.Printf("%s Command %s -> %s: %v\n", LengthRule, path, value, limit)
```
wtrocki, line 100: 1) I do not think we should print to standard out - stdnerr should be used for debug

validator/length.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	if length < limit.Min {
+		return Error{Name: fmt.Sprintf("length should be at least %d", limit.Min), Err: ErrMin, Rule: LengthRule}
```
wtrocki, line 104: This should he separate structure - not error

validator/must_present.go
```diff
@@ -0,0 +1,98 @@
+package validator
```
wtrocki, line 1: I will change package to rules

validator/length.go
ankithans: couldn't get you here?
Error is the separate struct which is defined as
```go
type Error struct {
	Name string
	Err  error
	Rule string
}
```

validator/must_present.go
```diff
@@ -0,0 +1,98 @@
...
+	"fmt"
+	"reflect"
+
+	"github.com/spf13/cobra"
+)
+
+// MustPresent is a struct that provides
+// Fields defined for MustPresent validation
+type MustPresent struct {
+	Fields []string
```
wtrocki, line 13: This will become map [string]Options once we get error levels etc.

validator/utils.go
```diff
@@ -0,0 +1,18 @@
...
+
+// Define rules as string
+var (
+	LengthRule      = "LENGTH_RULE"
+	MustPresentRule = "MUST_PRESENT_RULE"
+)
+
+// stats is used for providing info
+// about validation of commands
+type stats struct {
```
wtrocki, line 11: Capital letter and less generic name needed

validator/utils.go
```diff
@@ -0,0 +1,18 @@
+package validator
+
+// Define rules as string
+var (
+	LengthRule      = "LENGTH_RULE"
```
wtrocki, line 5: my preference would be that those will be part of the rule structure

validator/validator.go
```diff
@@ -0,0 +1,14 @@
+package validator
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// Rule is the interface which is implemented
+// by every rule defined in validator package
+type Rule interface {
+	// Validate validates the cobra command
```
wtrocki, line 10: Add name

ankithans: like different func name for each rule?

validator/must_present.go
```diff
@@ -0,0 +1,98 @@
...
+func (p *MustPresent) Validate(cmd *cobra.Command, verbose bool) []Error {
+	var errors []Error
+	info := stats{num: 0, num_failed: 0, errors: errors}
+
+	return p.ValidateHelper(cmd, verbose, info)
+}
+
+// ValidateHelper returns errors of type []Error
+// if the required field is not present
+func (p *MustPresent) ValidateHelper(cmd *cobra.Command, verbose bool, info stats) []Error {
```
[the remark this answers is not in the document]
ankithans: yes traversing through commands is same in all the rules

wtrocki:
We need to:
- format - unit tests vs ? 
- define usage and docs
- make it work for end users
- define how we distribute it - submodule etc.

I got panic when trying this in some CLI - will debug once we resolve other issues

validator/length.go
wtrocki: Too generic name. Lets change it

validator/error.go
wtrocki: define them.

Folder structure:
rules
   exist
        rule.go
        errros.go
        utils etc

validator/must_present.go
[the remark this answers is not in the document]
wtrocki: yes..

we can traverse outside and call rule for each command or have common traversing logic

validator/validator.go
wtrocki: Yes.. We need way to join text representation of rule with it's implementation. If that is not part of interface then there is no easy way to provide rule name as rules will be hardcoded inside lib.

wtrocki:
@ankithans while this resembles code that with some major refactor might work I would like to ask you to do quick 2 min excercise and draw class and sequence diagram on paper or some editor so we can model it better

wtrocki (changes requested):
needs documentation and clazz and sequence diagram

wtrocki:
rebase needed

ankithans:
Class diag

image

validator/error.go
ankithans: I think we can do this. As utils have same struct for all the rules. WDYT?
```
validator
     validator.go
     rules
          length.go
          exist.go
          utils.go
```

ankithans: And also some errors defined will be used by all rules, like ErrFieldNotExist, ErrUnsupported, ErrAbsent etc. So ig it's better to have errors defined commonly?

validator/length.go
ankithans: that was giving panic: reflect: call of reflect.Value.Type on zero Value

ankithans:
> I do not think we should print to standard out - stdnerr should be used for debug to get meaningful info (likrnjson) into std out.
> I think we should use logger :)

our logger is too basic 😅.
First we can add new functions in logger, then this can be resolved.
Can we do this in other PR. This has became too heavy. Even GitHub takes time to load it 😆

I have addressed the above comments. Needs a review @wtrocki

wtrocki:
> Can we do this in other PR

Please create issues for all extra elements that you do not feel comfortable to resolve in this PR.
That is pretty normal when starting.

> Can we do this in other PR. This has became too heavy. Even GitHub takes time to load it 😆

We will merge changes when ready. 

> This is left

You mean not right or not addressed? Lets create issues for those and work on them arfter this PR

> Class diag

As you see we have this anemic model where rules do a lot of responsibility. Ideally rules should do only rule validation. We need new issue with highest priority to address this by pulling all common elements into some ruleExecutor etc.

examples/plugins/date/main.go
```diff
@@ -49,5 +54,23 @@ func DateCommand() (*cobra.Command, error) {
 		},
 	}
 
+	// r2 := &rules.MustPresent{
```
wtrocki, line 57: Why commented out code.

examples/plugins/date/main.go
```diff
@@ -49,5 +54,23 @@ func DateCommand() (*cobra.Command, error) {
...
+	// 	Fields: []string{"Use", "Short", "Long", "Example", "SilenceUsage", "PreRun", "Hi"},
+	// }
+	// errr := r2.ValidateMustPresent(cmd, true)
+	// fmt.Println(errr)
+
+	ve := &rules.Length{
```
wtrocki, line 63: Why we have validator used in CLI?

validator/rules/traverse.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+// Traverse is used to traverse and validate
+// the command and it's descendant commands
+func Traverse(cmd *cobra.Command, verbose bool, info StatusLog, x interface{}, validate func(cmd *cobra.Command, verbose bool) []ValidationError) []ValidationError {
```
wtrocki, line 11: Wrong package. This is not part of the rules api

validator/rules/utils.go
```diff
@@ -0,0 +1,38 @@
+package rules
```
wtrocki, line 1: Should be part of the core instead of rules

validator/rules/utils.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"errors"
+
+	"github.com/spf13/cobra"
+)
+
+var (
+	ErrMin           = errors.New("less than min")
```
wtrocki, line 10: I had comment here that we should have those defined per rule

validator/rules/utils.go
wtrocki: utils is not good name for this

wtrocki:
@ankithans we not going to merge this PR without user facing api and docs. I'm asking for this for the 5th time (probably the reason why we have so many comments). Check discussion - this code is unusable. We cannot merge it even if 500 comments.

wtrocki:
We need to have user facing docs and proper API to merge.

ankithans:
@wtrocki yes completely fine and correct. So let's use this as testing library for cobra commands. 
So I should write unit tests for the validator? + User facing docs

validator/rules/utils.go
ankithans: I think we don't need this file, we can move it + traverse.go to validator.go

wtrocki: exactly!

docs/src/validator/validator_length_rule.md
```diff
@@ -0,0 +1,20 @@
+## Length Rule
+The lengths of properties in string format provided by `cobra.Command` can be controlled using the Length Rule.
+
+## How to use
+1. Create a length rule by providing the limits for each attribute you want to control
+```go
+myLengthRule := &rules.Length{
+    Limits: map[string]rules.Limit{
+        "Use":   {Min: 1, Max: 5},
```
wtrocki, line 9: Those values are not good default values

docs/src/validator/validator_length_rule.md
```diff
@@ -0,0 +1,20 @@
+## Length Rule
+The lengths of properties in string format provided by `cobra.Command` can be controlled using the Length Rule.
+
+## How to use
+1. Create a length rule by providing the limits for each attribute you want to control
+```go
```
wtrocki, line 6: Could you do some default validator that does initialize those values and allow users to override somethjng if needed.

docs/src/validator/validator_length_rule.md
```diff
@@ -0,0 +1,20 @@
+## Length Rule
```
wtrocki, line 1: As per our agreement for development this should be the first commit to this PR not the last

docs/src/validator/validator_length_rule.md
```diff
@@ -0,0 +1,20 @@
...
+myLengthRule := &rules.Length{
+    Limits: map[string]rules.Limit{
+        "Use":   {Min: 1, Max: 5},
+        "Short": {Min: 4, Max: 10},
+        "Long":  {Min: 5, Max: 20},
+        "Not":   {Min: 2, Max: 22},
+    },
+}
+```
+2. Use ValidateLength function to validate your cobra command with the rule created above. It takes a verbose bool for enabling/disabling debug StatusLogs
```
wtrocki, line 16: Each rule needs to be called explicitly?

My take is that end user api should finish in 2 lines no matter how many rules we have
```
validator := Validator{DefaultRuleConfig}
validator.vakidate(root)
```

Alternatively users should be able to override rule comfig.

wtrocki (changes requested):
We need to have example **how** rules can be used.
End user api should aggregate rules - no direct rule initialization.

docs/src/validator/validator_length_rule.md
ankithans: #### Now
```go
var lenValidator rules.Length // user can give their config here or use default like this
lenValidator.ValidateLength(cmd, true)
```
Is the above fine?
> Each rule needs to be called explicitly?

As rule is a separate package now, we might be able to figure out something when we do ruleExecutor

ankithans:
testing of cmd
https://github.com/aerogear/charmil/tree/cobra-validator/validator/example

docs
https://github.com/aerogear/charmil/blob/cobra-validator/docs/src/validator/validator.md

validator/rules/executor.go
```diff
@@ -0,0 +1,89 @@
...
+func (config *RuleConfig) initDefaultRules() {
+
+	// default config for rules
+	var defaultConfig = &RuleConfig{
+		Verbose: false,
+		Length: Length{
+			Limits: map[string]Limit{
+				"Use":     {Min: 2, Max: 20},
+				"Short":   {Min: 15, Max: 200},
+				"Long":    {Min: 100, Max: 10000},
```
wtrocki, line 56: do not put max by default

ankithans: default max = `math.MaxInt32`
this would be fine?

wtrocki: No.. No max present

wtrocki:
Rebase needed

docs/src/validator/validator.md
```diff
@@ -0,0 +1,37 @@
...
+## How to use
+It is recommended to use the validator while writing unit tests for cobra commands.
+
+1. Create a validator of type `rules.RuleConfig`. You can provide your own RulesConfig or use the default one by leaving it empty
+```go
+var vali rules.RuleConfig
+```
+or overriding default config
+```go
+var vali rules.RuleConfig = rules.RuleConfig{
```
wtrocki, line 19:
```suggestion
var vali := rules.RuleConfig{
```

validator/example/cmd_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"io/ioutil"
+	"testing"
+
+	"github.com/aerogear/charmil/validator/rules"
+)
+
+func Test_ExecuteCommand(t *testing.T) {
+	cmd := NewCommand()
+
+	// Testing cobra commands with default recommended config
```
wtrocki, line 14: Remove this! No commented code on master

validator/example/cmd_test.go
```diff
@@ -0,0 +1,39 @@
...
+		var vali rules.RuleConfig = rules.RuleConfig{
+				Verbose: true,
+				MustExist: rules.MustExist{
+					Fields: []string{"Args"},
+				},
+			}
+	*/
+	var vali rules.RuleConfig
+	validationErr := vali.ExecuteRules(cmd)
+	for _, errs := range validationErr {
```
wtrocki, line 26: This is not test :) 
Test does asertions :)
We should at least check if no errors are returned.
We can also print out some info etc.

THis test should be passing as we going to use it for CI/CD #93

validator/example/cmd_test.go
```diff
@@ -0,0 +1,39 @@
...
+		if errs.Err != nil {
+			t.Errorf("%s: cmd %s: %s", errs.Rule, errs.Cmd.CommandPath(), errs.Name)
+		}
+	}
+
+	b := bytes.NewBufferString("")
```
wtrocki, line 32: What is this chunk of code doing? Some leftover/unfinished idea?

wtrocki (approved):
This is ok to merge.

https://gfycat.com/athleticsaneaztecant-wolf-of-wall-street-leonardo-dicaprio

ankithans:
@wtrocki addressed above comments

wtrocki:
> Rebase needed
> This branch cannot be rebased due to conflicts

One comment missed

ankithans:
done @wtrocki

wtrocki:
Merged with 101 comments:
https://en.wikipedia.org/wiki/101_(topic)

ankithans:
"one-oh-one" 😅🥲

wtrocki:
Btw.. This is pretty much still work in progress change in main. Let's work on refining it better.

ankithans:
yes sure, first let's do unit tests properly and find bugs by trying with big CLI's
first #88 #94
then #84 ,#85, #86

then we should do CLI? or in between what do you suggest?

wtrocki:
Priority nr1 is to do rhoas integration.

jackdelahunt:
@ankithans I just worked with the validator functionality for the first and it looks really nice just had a few questions on going forward on what we are looking to refine.

- We check the rules config values to make sure they are not dumb, like is max less then min etc. [here](https://github.com/aerogear/charmil/blob/ade49e963aba170f8187baecc0443ca9a10beed1/validator/rules/length.go#L96) but I don't see why we can't check these fields when we create the rule config instead of every time we run the validator. 
- Also cases like if max is 0 then don't check should be documented but again you may already be planning this. [Code](https://github.com/aerogear/charmil/blob/ade49e963aba170f8187baecc0443ca9a10beed1/validator/rules/length.go#L87)

ankithans:
> We check the rules config values to make sure they are not dumb, like is max less then min etc. here but I don't see why we can't check these fields when we create the rule config instead of every time we run the validator.

I think, it's better to be part of rule, which can be run by ruleExecutor. As every rule will have some checking like this. So if we do that in ruleExecutor, things might get messed up. WDYS?

> Also cases like if max is 0 then don't check should be documented but again you may already be planning this. Code

Yep thanks for suggesting, need to document again

jackdelahunt:
> if we do that in ruleExecutor, things might get messed up.

Yes that being in the validator would be annoying but I don't think it has to be. 

I am gonna make a quick pr today on what think we can do as I think I have a good idea. We also have to be mindful of yaml/json configs for the rules. But I will get back to you.

PR #83, opened 2021-06-24
Add CI/CD & issue pr templates

### LINKED ISSUE #53: Setup CI/CD to the repository
For examples see: https://github.com/redhat-developer/app-services-sdk-go
makefile should define steps

ankithans:
closes #53 
I thought, it would be beneficial to have from start.

@aerogear/charmil-core


wtrocki:
I would much preffer to actually use makefile for this rather than have hardcoded flows in workflow. 

Let's focus on nontrivial work and try to push core changes. 
Leave some simple work to others/later as without core there is not much sense to invest. 

If there is need for some automation let me know. We have standards for makefiles etc. so it can be copy paste of workflows etc.

Issue #69, opened 2021-06-17
Configuration storage and abstraction for plugin


ankithans:
For a clean default configuration, yaml files can be used. [viper](https://github.com/spf13/viper) might be useful here which is used by cobra as well!

wtrocki:
Ther the is no such thing as default configuration.
Basically plugin commandline can exist on it's own and have their own config. 
If we include that in host CLI we ending up with 2 configs - not good. 

We need to have a way to compose config and share some common parts. 
This is blocked by figuring out authentication to the CLI etc.

PR #87, opened 2021-06-26
feat: add dev preview feature

### LINKED ISSUE #72: Provide feature flags to enable specific plugins
Generally, feature flags allow us to activate or deactivate specific features of our application. 
In our case, we'll be allowing the SDK users to use feature flags to activate certain plugins.

This requires further investigation and discussion before starting out.

namit-chandwani:
Closes #72 
Based on discussion: https://github.com/aerogear/charmil/discussions/81

Still WIP.

## TODO
- [x] ~~Add inline documentation~~
- [x] ~~Add external docs~~
- [ ] Internationalize all strings
- [ ] Replace fmt.Println() statements with suitable logger functions
- [ ] Add tests
- [ ] Replace `.env` config storage approach with a better one, once issue: [#69] is resolved
---
## `config.env` file content:
```env
CHARMIL_DEV_PREVIEW_ENABLED="true" (or) "false"
```

### DIFF
--- a/examples/plugins/echo/locals/en/echo.en.yaml
+++ b/examples/plugins/echo/locals/en/echo.en.yaml
@@ -1,0 +1,23 @@
+echo.cmd.use:
+  description: "The one-line usage message"
+  one: "echo [string(s) to echo]"
+
+echo.cmd.short:
+  description: "The short description shown in the 'help' output"
+  one: "Prints given strings to stdout"
+
+echo.cmd.example:
+  description: "Examples of how to use the command"
+  one: "$ host echo Hey Gophers"
+
+echo.cmd.flag.repeat.description:
+  description: "Repeat flag description"
+  one: "number of times to echo to stdout"
+
+echo.cmd.flag.uppercase.description:
+  description: "Uppercase flag description"
+  one: "convert to uppercase"
+
+echo.cmd.flag.lowercase.description:
+  description: "Lowercase flag description"
+  one: "convert to lowercase"
--- a/examples/plugins/echo/main.go
+++ b/examples/plugins/echo/main.go
@@ -10,25 +10,37 @@ import (
 	"golang.org/x/text/language"
 )
 
+// EchoCommand returns echo plugin's root command struct.
+// This will be added to the host CLI as an extension.
 func EchoCommand() (*cobra.Command, error) {
 	var (
-		iter                 int
-		uppercase, lowercase bool
+		// Stores the number of iterations specified through the `repeat` flag.
+		iter int
+
+		// Holds a `true` value if the uppercase flag is set.
+		uppercase bool
+
+		// Holds a `true` value if the lowercase flag is set.
+		lowercase bool
 	)
 
+	// Stores the config for localizer
 	locConfig := localize.Config{
 		Language: language.English,
 		Path:     "examples/plugins/echo/locals/en/echo.en.yaml",
 		Format:   "yaml",
 	}
 
+	// Initializes the localizer by passing config
 	loc, err := localize.InitLocalizer(locConfig)
 	if err != nil {
 		return nil, err
 	}
 
+	// Creates a new factory instance using default configuration
 	f := factory.Default(loc)
 
+	// Stores the root command of plugin
 	echoCmd := &cobra.Command{
 		Use:          f.Localizer.LocalizeByID("echo.cmd.use"),
 		Short:        f.Localizer.LocalizeByID("echo.cmd.short"),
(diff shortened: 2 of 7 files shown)


examples/plugins/echo/locals/en/echo.en.yaml
```diff
@@ -0,0 +1,23 @@
+echo.cmd.use:
+  description: "The one-line usage message"
+  one: "echo [string(s) to echo]"
+
+echo.cmd.short:
+  description: "The short description shown in the 'help' output"
+  one: "Prints given strings to stdout"
+
+echo.cmd.example:
```
wtrocki, line 9: why we need this

examples/plugins/echo/main.go
```diff
@@ -0,0 +1,63 @@
...
+	echoCmd := &cobra.Command{
+		Use:          f.Localizer.LocalizeByID("echo.cmd.use"),
+		Short:        f.Localizer.LocalizeByID("echo.cmd.short"),
+		Example:      f.Localizer.LocalizeByID("echo.cmd.example"),
+		SilenceUsage: true,
+		Args:         cobra.MinimumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			message := strings.Join(args, " ")
+
+			if uppercase {
```
wtrocki, line 41: why this logic?

pkg/preview/preview.go
```diff
@@ -0,0 +1,71 @@
...
+	"os"
+	"strconv"
+
+	"github.com/fatih/color"
+	"github.com/joho/godotenv"
+	"github.com/spf13/cobra"
+)
+
+type devPreview struct {
+	commands  []*cobra.Command
```
wtrocki, line 14: what are commands for

pkg/preview/preview.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	dp.isEnabled = boolVal
+}
+
+func UpdateFlagValue(cmd *cobra.Command, args []string) {
+	if cmd.Flags().Changed("dev_preview") {
+		env, err := godotenv.Unmarshal("CHARMIL_DEV_PREVIEW_ENABLED=" + strconv.FormatBool(dp.isEnabled))
```
wtrocki, line 41: Lets not write code we do not plan merging

wtrocki:
@namit-chandwani this breaks contract we agreed on that we focus on usage and docunention frist

namit-chandwani:
> @namit-chandwani twe agreed on that we focus on usage and docunention frist. Do you mind adding dev usage docs. it is hard to review this.

Sure, I'll add them and ping you once done

examples/plugins/echo/main.go
namit-chandwani: That's just a part of an example plugin that I've added and doesn't do much.
I wanted to have a basic and minimal single command example plugin, so I chose to add this. 😅

All it does is clone the functionality of the Linux `echo` command ie. print all the strings passed as arguments, into stdout.
It has 3 local flags:
- **uppercase**: If this is enabled, strings are converted to uppercase before printing into stdout
- **lowercase**: If this is enabled, strings are converted to lowercase before printing into stdout
- **repeat**: The value that is passed into this flag is the number of times the input strings will be printed to stdout

pkg/preview/preview.go
namit-chandwani: `commands` stores all the commands that have been added for dev preview by the host CLI developer.

I've added comments for the same in the file now, please have a look.

wtrocki:
lets leave this PR for now. It is not mergable unti.we have config anyway

namit-chandwani:
> lets leave this PR for now. It is not mergable unti.we have config anyway

Agreed. Will start working on the config issue now and come back to this PR only once that's done

examples/plugins/echo/main.go
wtrocki: As we discussed I think it is best for examples to be minimal.
Less non functional code then less things that people need to read

pkg/preview/preview.go
```diff
@@ -0,0 +1,93 @@
...
+	for _, cmd := range dp.commands {
+		updateCmdFields(cmd)
+	}
+}
+
+// updateCmdFields modifies fields of the command passed in
+// as argument with appropriate values to mark it as preview.
+func updateCmdFields(cmd *cobra.Command) {
+	cmd.Annotations = map[string]string{"channel": "preview"}
+	cmd.Short = color.HiCyanString("[Preview] ") + cmd.Short
```
wtrocki, line 83: Devs would probably want this configurable

wtrocki (changes requested):
I think you got this rhoas inspired code right but we need to keep it short and do less opiniated and non functional code.

wtrocki:
I did remove this feature from rhoas as it wasn't really well done.

namit-chandwani:
> Not sure how to drive that from now on.

If it's fine with you, I can try coming up with a different approach to achieve this.
As there's already a config management system in place now, there won't be any blockage to this PR like it was before

> I did remove this feature from rhoas as it wasn't really well done.

Can you please let me know what were the things that you found wrong with the Rhoas implementation?
I'll try to avoid those in my approach.

wtrocki:
On top of pririoties I will leave that PR for now and focus on validation of config in the starter that were merged.

namit-chandwani:
> On top of pririoties I will leave that PR for now and focus on validation of config in the starter that were merged.

Got it. Will switch my focus to that

ankithans:
shall we close it @namit-chandwani ?

namit-chandwani:
> shall we close it @namit-chandwani , while keeping the branch

Sure, closing it for now. 
Will open it back when needed

Issue #90, opened 2021-06-28
Use command validator as CLI


ankithans:
Currently we are using validator as Unit test framework. While writing tests for a cobra command, `ExecuteRules` function can be executed which will return a slice of `ValidationErrors`.
But as most of the users will prefer a CLI tool for this, there will some significant changes to be made in validator API, which are yet to be explored!

All the `RuleConfig` can be loaded from a json/yaml file into the CLI.

(https://github.com/aerogear/charmil/issues/89#issuecomment-869746695) 
If we make CLI for validating commands then how will we provide cobra command from a file to the CLI. Read the file?

jackdelahunt:
Can you layout the issue a bit more here. I think I know what you mean but I am not sure.

How I see the commands validator as a cli is you build with all the commands you want to add and the rules you setup with a config file. Then from that build we run `charmil validate` or something like that. This then runs each rule as defined by us or by the config file on each commanded added.

wtrocki:
Thank you @jackdelahunt for keeping eye on issues and validation. This is hugely benefitial for students and brings diversity.

> How I see the commands validator as a cli is you build with all the commands you want to add and the rules you setup with a config file. Then from that build we run charmil validate or something like that. This then runs each rule as defined by us or by the config file on each command added.

This is 100% correct with what I was thinking.
We both know that this will require more work/steps comparing to unit tests approach.

ankithans:
> Can you layout the issue a bit more here. I think I know what you mean but I am not sure.

updated, thanks for pointing out!

> How I see the commands validator as a cli is you build with all the commands you want to add and the rules you setup with a config file. Then from that build we run charmil validate or something like that. This then runs each rule as defined by us or by the config file on each command added.

okay, we can use binary of a CLI(which we want to validate), and pass that binary into `charmil validator CLI`. Then how we will access the attributes of cobra.Command to validate(like Use, Short etc). Binary will only allow use to use the CLI.
or I am taking `build` in wrong sense?

wtrocki:
> okay, we can use binary of a CLI(which we want to validate), and pass that binary into charmil validator CLI

We will need to compile our code + cobra commands into executable, but maybe there is some golang trick to build tool that does something at compile time (probably not)

jackdelahunt:
@wtrocki and @ankithans is it okay if I tackle this issue? If this was expected to be done by the students that I am okay to leave it to them.

wtrocki:
[...]

PR #96, opened 2021-06-29
Validate commands through the cli of the host

jackdelahunt:
### Description
Find discussion -> #90 

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

## Screenshots (if appropriate):


docs/src/charmil_architecture.md
```diff
@@ -0,0 +1,26 @@
...
+To run commands that are defined by the `ack` team you use the standard syntax for calling a command.
+```bash
+> ack print Hello
+Hello
+```
+To run commands that are defined by the `charmil` team you use the charmil command and then any sub-command you want.
+```bash
+> ack charmil validate
+All custom commands are valid
+```
```
ankithans, line 20: so Charmil is a standalone CLI added into the host `ack`

docs/src/charmil_architecture.md
```diff
@@ -0,0 +1,26 @@
...
+
+### Disabling the charmil provider
+Because `charmil` commands are made for the host and not for the end user you can disable them being available for a release build. 
+```bash
+> go build ack --CHARMIL_COMMANDS=false
+```
```
ankithans, line 26: I would say great explanation with the example!

docs/src/charmil_architecture.md
ankithans: If we do this, do we need input processor?

jackdelahunt: What I mean by input processor is just parsing the input from the cli into the parts we want. Cobra might does this already for us and if not there is definitely something out there.

docs/src/charmil_architecture.md
```diff
@@ -0,0 +1,26 @@
...
+To run commands that are defined by the `charmil` team you use the charmil command and then any sub-command you want.
+```bash
+> ack charmil validate
+All custom commands are valid
+```
+
+### Disabling the charmil provider
+Because `charmil` commands are made for the host and not for the end user you can disable them being available for a release build. 
+```bash
+> go build ack --CHARMIL_COMMANDS=false
```
wtrocki, line 25: I would avoid that. Lets build separate cli for testing things

jackdelahunt: Updated the docs, rethought how we want to add our commands. In the docs I stated that they are opt-in and have to specifically built for.

jackdelahunt:
To address how I got the conditional building to work, I first tried to look for features in go that are similar to the conditional compilation tools in c/c++. Because go doesn't have a pre-processor only build tags can be used to specify which files go in the build. This is why the files `enable_commands.go` and `disable_commands.go` are added.

Makefile
```diff
@@ -1,8 +1,13 @@
+TAGS="dev"
+
 run:
 	go run ./examples/host
 
 build:
 	go build ./examples/host
 
+build/tags:
+	go build -tags ${TAGS} ./examples/host
```
wtrocki, line 10: Does this means that we no longer library and we will provide some form of CLI as part of release?

core/commands/validate.go
```diff
@@ -0,0 +1,25 @@
...
+func ValidateCommand() (*cobra.Command, error) {
+	// creating new command
+	// using localizer to access default text by ID provided in locals
+	cmd := &cobra.Command{
+		Use:     "validate",
+		Short:   "validate custom commands",
+		Long:    "validates the custom commands with the defined rules",
+		Example: "host charmil validate",
+		Args:    cobra.ExactArgs(0),
+		// SilenceUsage: true,
```
wtrocki, line 17: No comments

core/commands/validate.go
```diff
@@ -0,0 +1,25 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			println("validate")
```
wtrocki, line 19: Should something be here?

core/commands/validate.go
```diff
@@ -0,0 +1,25 @@
+package commands
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// Date Command
+func ValidateCommand() (*cobra.Command, error) {
```
wtrocki, line 8: I like idea of command that people can add and build based on their own root command. Trick is that this command would probably be root command on it's own. Having validate command accepting your root command will be dependency cycle :)

docs/charmil_commands.md
```diff
@@ -0,0 +1,26 @@
+## Charmil Commands
+Charmil provides prebuilt commands for developers to use when testing their application. These commands are opt-in and are only added when the application is built with a specific tag.
```
wtrocki, line 2: I like idea of helper commands that users can do.
Good example would be config command that @namit-chandwani is working.

I'm actually working on this issue now that does common config changes:
https://github.com/redhat-developer/app-services-cli/issues/346

docs/charmil_commands.md
```diff
@@ -0,0 +1,26 @@
...
+
+To add charmil commands to your root command you can call the `CharmilCommands()` function. If calling this function does not populate `err` then add the commands to your root command.
```
wtrocki, line 4: I love this docs however we do not have clear example for the commands like this right now or CharmilCommands() in core. Thinking how to keep this work so we can actually implement things this way.

One of the ideas would be to push this docs to master in separate PR and create placeholder for charmil commands. 
Then rest of the team will simply reuse that placeholder to inject commands as they build features. 
This will be amazing for us to have docs and end user API first and will solve challenges on "HOW TO USE THIS DAM THING :) "

wtrocki:
@jackdelahunt I feel that this PR does 2 amazing things:

- It exposes validate as helper and tries to provide some example how to use it as binary (although this still needs some refinement
- It brings amazing documentation and good development approach that we can employ for charmil to produce value. 


My take is that this PR is actually 2 separate PRs. If you look at individual comments it will be clear what I mean

core/commands/validate.go
jackdelahunt: Yes this is just a proof of concept to get the charmil commands working.

Makefile
jackdelahunt: I would say still a library with an optional cli if a developer wants it.

core/commands/validate.go
jackdelahunt: I see alot of comments that I left in from copying the date command, will clean this up.

core/commands/validate.go
wtrocki: Yes.. but docs are actually quite good example how things might work. If we split we can merge docs and leave validation POC separate.

PR #103, opened 2021-07-01
fix: validate rules in loop

### LINKED ISSUE #53: Setup CI/CD to the repository

### LINKED ISSUE #93: CI/CD support for validator
Run tests and compile binary with PR

### LINKED ISSUE #98: Rule methods should implement interface
To make it easy for adding more custom rules to validator, rules methods should implement an interface. 
we should have the same interface and the same function. Without that there will be no way to add rule to validator. This two methods should be the same method

### Currently
lengthRule implements - `validateLength`
MustExistRule implements - `validateMustExist`

ankithans:
Closes #98 
Closes #93 
Closes #53 

### Description
- Rules implement `Rule` interface
- Validate rules in for loop(DRY), instead of calling one by one
- CI/CD tests & builds validator example

### Type of change

### DIFF
--- a/validator/rules/executor.go
+++ b/validator/rules/executor.go
@@ -10,6 +10,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type Rules interface {
+	Validate() []validator.ValidationError
+}
+
 // RuleConfig is the struct that stores
 // configuration of rules
 type RuleConfig struct {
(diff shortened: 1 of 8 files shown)


validator/rules/executor.go
```diff
@@ -10,6 +10,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type Rules interface {
```
wtrocki, line 13: doc

validator/rules/executor.go
```diff
@@ -10,6 +10,10 @@ import (
...
+	Validate() []validator.ValidationError
```
wtrocki, line 14: Wrap error aray in result structure for other elements

wtrocki (approved):

wtrocki:
Lets release this /create github release v0.1.0

validator/rules/executor.go
wtrocki: actually lets leave it likt thid

ankithans:
Draft release. Changes required?
https://github.com/aerogear/charmil/releases/tag/untagged-5c6b0883d263545a9332

PR #105, opened 2021-07-01
refactor RuleConfig + overriding default values

### LINKED ISSUE #104: Cannot override default rules
## Description
[Priority High] Rules cannot be overridden using RuleConfig. Below code uses default config only.
```go
var vali rules.RuleConfig = rules.RuleConfig{
	Length: rules.Length{
		Limits: map[string]rules.Limit{
			"Example": {Min: 10, Max: 100},
		},
	},
}
```

ankithans:
Closes #104 

### Description
Now Rule Config can be overridden by the user 
- [x] refactor ruleConfig and rules interface
- [x] overriding of rules by the user

### Type of change

- [x] Documentation change


validator/rules/executor.go
```diff
@@ -113,28 +116,22 @@ func (config *RuleConfig) initDefaultRules() {
 		},
 	}
 
-	// Check verbose input from user
-	var verbose bool
-	if config != nil && config.Verbose {
-		verbose = true
+	if err := mergo.Merge(&config.Length, defaultConfig.Length); err != nil {
```
wtrocki, line 119: Why we would merge specific rules. Can we merge all config paths?

wtrocki, line 119: I think overal config is confusing. WE should either:

- Add rules field with array of IRule
- Put string Rule after each rule(not like it personally)

ankithans: okay, so I really need to rethink the architecture, as everything is dependent on config. I will refactor it, keeping the functionality same!

I have one more concern - here we don't really need validator package, shall we do everything under rules package? But this can be changed while doing above refactoring!

wtrocki: Yes.. No pressure. Take it easy

dimakis: If the aim here is to merge two config files into one giving precedence to a passed in file but keeping default values were no values are passed in and we are already using `mergo`. We could try using merge with the transformer: `mergo.Merge(&config.Length, defaultConfig.Length, mergo.WithOverride)`. This should keep config with first values whilst only updating fields passed in from `defaultConfig.Length`.

Speaking of `.Length` is this the best name for this field? is it a validated ( element size wise ) map of config fields? I may just not be getting it either, in which case my suggestion may not be applicable.

dimakis (commented):
great work @ankithans :+1:

validator/rules/executor.go
ankithans: I got you point @dimakis. But we want to keep defaultConfig.Length as it is and merge config.Length into it. We can do opposite as you suggested but then we need to do `*config = *defaultConfig`, extra lines of code. This can be avoided I'd say.

> Speaking of .Length is this the best name for this field? is it a validated ( element size wise ) map of config fields? I may just not be getting it either, in which case my suggestion may not be applicable.

Length is a struct of Limits (that is map[string]Limit) and Limit is struct consisting of Max, Min length of field(int value). We can think over the names again.

> Why we would merge specific rules. Can we merge all config paths?

Problem was mergo cannot merge 2 slices. It can only merge structs and maps. And the previous method also had similar problems.

dimakis: > I got you point @dimakis. But we want to keep defaultConfig.Length as it is and merge config.Length into it. We can do opposite as you suggested but then we need to do `*config = *defaultConfig`, extra lines of code. This can be avoided I'd say. And btw this code might change as there is some major refactoring going on.
> It may be one extra assignment, but it negates the need to check for duplication of values in a slice -> `removeDuplicates()` becomes obsolete.
> > Speaking of .Length is this the best name for this field? is it a validated ( element size wise ) map of config fields? I may just not be getting it either, in which case my suggestion may not be applicable.
> 
> Length is a struct of Limits (that is map[string]Limit) and Limit is struct consisting of Max, Min length of field(int value). We can think over the names again.
>  +1

wtrocki:
Is that ready for review?

ankithans:
Not yet, I will ping you, fixing an issue after refactor.

ankithans:
One thing left in this PR ie. overriding. I am stuck and trying to figure out the way to do this! Mergo cannot be used.
#### Merge this
```go
RuleConfig{
	Rules: []rules.Rules{
		&rules.Length{Verbose: true, Limits: map[string]rules.Limit{"Use": {Min: 100}}},
	},
}
```
#### into this
```go
RuleConfig{
	Rules: []Rules{
		&Length{
			Verbose: false,
			Limits: map[string]Limit{
				"Use":     {Min: 2},
				"Short":   {Min: 15},
				"Long":    {Min: 50},
				"Example": {Min: 50},
			}},
		&MustExist{Verbose: false, Fields: []string{"Use", "Short", "Long", "Example"}},
	},
}
```
#### Result should be
```go
RuleConfig{
	Rules: []Rules{
		&Length{
			Verbose: true,
			Limits: map[string]Limit{
				"Use":     {Min: 100},
				"Short":   {Min: 15},
				"Long":    {Min: 50},
				"Example": {Min: 50},
			}},
		&MustExist{Verbose: false, Fields: []string{"Use", "Short", "Long", "Example"}},
	},
}
```

dimakis:
@ankithans do you have an example of how you are trying to use it? can you push it so i can have a look?

ankithans:
@dimakis this is the example I am using for testing https://github.com/aerogear/charmil/tree/ankithans/issue104/validator/example

validator/rules/executor.go
```diff
@@ -95,46 +91,23 @@ func (config *RuleConfig) validate(cmd *cobra.Command, info *validator.StatusLog
...
-			log.Fatal(err)
-		}
-		data := []byte(out)
-
-		errr := json.Unmarshal(data, &defaultConfig)
-		if errr != nil {
-			log.Fatal(errr)
-		}
-	}
+	// TODO: Override default configuration
```
dimakis, line 110: @ankithans All I'm seeing is an empty TODO, have/ can you push up your implementation?

ankithans: sure give me sometime, I will ping you once done. i want to push some working code.

dimakis: It's not essential for it to be working, just change the PR to a [WIP]. I enjoy that you don't want to push non functional code, but if you are still facing issues push it and I'll see if I can help. Don't get too hung up on a single piece of the puzzle. 👍

ankithans: thank you so much!!!
Now Length rule is working fine! as it has a map field Limits to merge
but MustExist cannot be merged with mergo as it is a slice. One solution is convert 
```go
Fields []string
```
to
```go
Fields map[string]bool
```

ankithans:
@wtrocki @dimakis ready for review

validator/rules/executor.go
ankithans: I did it this way, changed the Fields []string to Fields map[string]bool. This way user can disable the default MustExist fields(which was not possible previously)

wtrocki:
@ankithans configuration and defaults are common problem. Please reseach this topic first.

1 min Google returned me this:
https://github.com/containous/staert

Although I'm not fan of libraries

Please check how golangci lint does it
https://github.com/golangci/golangci-lint

wtrocki:
Good start is to query google properly:
Example:

https://www.google.com/search?q=golang+struct+default+value&client=ms-android-xiaomi-rvo3&ei=jkHgYLm4CYeT8gL4p5zQAQ&oq=golang+structure+defaults&gs_lcp=ChNtb2JpbGUtZ3dzLXdpei1zZXJwEAEYADIICAAQFhAKEB46BAgAEEc6AggpOgUIABCRAjoCCAA6BAgAEA06BggAEBYQHjoFCCEQoAFQiE9Y2HxgxYYBaANwAXgAgAFqiAHMEZIBBDMyLjGYAQCgAQHIAQjAAQE&sclient=mobile-gws-wiz-serp

ankithans:
@wtrocki How I did this now is according to #110 
User API looks like this, as you also and @jackdelahunt suggested.
```go
vali := rules.RuleConfig{
	Verbose: false,
	Rules: []rules.Rules{
		&rules.Length{Limits: map[string]rules.Limit{"Use": {Min: 2}}},
		&rules.MustExist{Fields: map[string]bool{"Long": false}},
	},
}
```
And we have solved the merging of the default config & user provided config in efficient way, using mergo 
https://github.com/aerogear/charmil/blob/ankithans%2Fissue104/validator/rules/executor.go#L96
```go
if err := mergo.Merge(&defaultConfig, config, mergo.WithSliceDeepCopy); err != nil {
	log.Fatal(err)
}
*config = defaultConfig
```
seen golangci-lint configuration file, it also accepts diff config for each linter.
Also got much help from this issue opened in mergo a long time back https://github.com/imdario/mergo/issues/121 Many were facing the same issue as I was!

wtrocki:
I think problem is that we trying doing something different:

We doing merge that is tricky in go, but can do default values for structure which is supported by go core (with json) or covered by some libs

ankithans:
Okay, I got your point now! But will it work with slice? need to initialize slice of Rule Interface.
How will it work with Min in length, as that will be different for each attribute in cobra.
I think it will be better to initialize manually in the array of rules, what do you think?

wtrocki:
Just have all as structure. 

This structure like we have unneded complexity/slices etc.

Have structure that is parsable to json/yaml
```
vali := rules.RuleConfig{
	Verbose: false,
	Rules: []rules.Rules{
		&rules.Length{Limits: map[string]rules.Limit{"Use": {Min: 2}}},
		&rules.MustExist{Fields: map[string]bool{"Long": false}},
	},
}
```

ankithans:
> Witten on phone while feeding baby. Sorry for typos

No issues 😄

```
ValidatorConfig{
 RuntimeOptions,
Rules:{
 LenghtRule, 
 MustExist
}
}
LenghtRule{
  Max int32
  Min int32
}
```
If we follow above struct we would need to add rules in ruleExecutor manually, and that was what we wanted to solve. For maintainability we did choose array of IRules. Previously we were doing this only, accepting struct...

wtrocki:
> If we follow above struct we would need to add rules in ruleExecutor manually

no. i keep repating it so will explain like we are both five y old.

Config is like piece of paper that tells us about the shoping list. executor is send to shop. he picks list and starts collecting goods (rules) to execute/buy.

Config should not be used in rule executor to call validate. it should be different structure - rule itself. I mentioned it coupke times but did not explained it.

ankithans:
> typed with one hand while walking, sorry for typos

i really can't thank you enough for these explanations. I will work on it tomorrow

ankithans:
@wtrocki review required

validator/example/cmd_test.go
```diff
@@ -11,15 +11,24 @@ func Test_ExecuteCommand(t *testing.T) {
 
 	// Testing cobra commands with default recommended config
 	// default config can also be overrided
-	var vali rules.RuleConfig
-	validationErr := vali.ExecuteRules(cmd)
+	ruleCfg := rules.ValidatorConfig{
+		Options: rules.Options{Verbose: false},
```
wtrocki, line 15: Too Generic name. Options of what?

ankithans: done! golangci-lint failing

wtrocki:
Rebase will solve this :D

validator/example/cmd_test.go
wtrocki: @ankithans I think fixing linting rules will be separate issue - good first issue

validator/rules/config.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/imdario/mergo"
+	"github.com/spf13/cobra"
+)
+
+// ValidatorConfig is provided to user for overriding default rules
+type ValidatorConfig struct {
+	ValidatorOptions `json:"ValidatorOptions"`
+	ValidatorRules   `json:"ValidatorRules"`
+}
+type ValidatorOptions struct {
```
wtrocki, line 17: Missing docs

validator/rules/config.go
```diff
@@ -0,0 +1,72 @@
...
+// provided by validatorConfig
+func (validatorConfig *ValidatorConfig) ExecuteRules(cmd *cobra.Command) []validator.ValidationError {
+	var ruleConfig RuleConfig
+	return ruleConfig.ExecuteRulesInternal(cmd, validatorConfig)
+}
+
+func ValidatorConfigToRuleConfig(validatorConfig *ValidatorConfig, ruleConfig *RuleConfig) {
+	defaultVerbose := validatorConfig.ValidatorOptions.Verbose
+
+	defaultConfigJson := `{
```
wtrocki, line 35: Wrong place to put default config :)

validator/rules/config.go
```diff
@@ -0,0 +1,72 @@
...
+		"ValidatorOptions": {
+			"Verbose": false
+		},
+		"ValidatorRules": {
```
wtrocki, line 39: The structure will be much easier to maintain

validator/rules/config.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+	}`
+
+	// unmarshal defaultConfigJson in configHelper
+	var configHelper ValidatorConfig
+	if err := json.Unmarshal([]byte(defaultConfigJson), &configHelper); err != nil {
+		log.Fatal(err)
+	}
+
+	configHelper.Length.Verbose = defaultVerbose
```
wtrocki, line 60: What is that about?
Very bad smell in code. Direct assignment. Why not make rule working with the validationOptions?

validator/rules/config.go
```diff
@@ -0,0 +1,72 @@
...
+	configHelper.MustExist.Verbose = defaultVerbose
+
+	// Merge user provided config into configHelper
+	if err := mergo.Merge(&configHelper, validatorConfig, mergo.WithSliceDeepCopy); err != nil {
```
wtrocki, line 64: Do we still need to use mergo?

validator/rules/executor.go
```diff
@@ -1,40 +1,38 @@
...
-	MustExist
+	Rules []Rules
 }
 
-// ExecuteRules executes all the rules
-// according to the RuleConfig provided
-func (config *RuleConfig) ExecuteRules(cmd *cobra.Command) []validator.ValidationError {
+// ExecuteRulesInternal executes all the rules
+// provided by ruleConfig
+func (ruleConfig *RuleConfig) ExecuteRulesInternal(cmd *cobra.Command, userValidatorConfig *ValidatorConfig) []validator.ValidationError {
```
wtrocki, line 25: Do you dissagree with what I have proposed? 
In what I have proposed config should not have any method (not to mention internal)

wtrocki (changes requested):
This somewhat ended up being more convoluted than before. 
Do you have any problems with what was proposed? Any challenges?

validator/rules/config.go
ankithans: yes..
`default = Fields{"Use": true, "Short": true, "Long": true, "Example": true}`
if user specified `Fields: {"Args": true, "Long": false}`, 
then without mergo it will be - `Fields: {"Args": true, "Long": false}`

but it should be  `Fields{"Use": true, "Short": true, "Long": false, "Example": true, "Args": true}`

validator/rules/executor.go
ankithans: @wtrocki addressed comments

wtrocki:
@ankithans would it be possible to update the documentation and example code?

validator/example/cmd_test.go
```diff
@@ -11,15 +11,24 @@ func Test_ExecuteCommand(t *testing.T) {
 
 	// Testing cobra commands with default recommended config
 	// default config can also be overrided
-	var vali rules.RuleConfig
-	validationErr := vali.ExecuteRules(cmd)
+	ruleCfg := rules.ValidatorConfig{
+		ValidatorOptions: rules.ValidatorOptions{Verbose: false},
```
wtrocki, line 15: Do we need to specify value that is used by default?

```suggestion
```

ankithans: @wtrocki review required

wtrocki: Of the whole PR? or just this?

wtrocki (approved):

validator/example/cmd_test.go
ankithans: whole

ankithans: now merged

wtrocki: 🙈

PR #106, opened 2021-07-01
Charmil Commands

jackdelahunt:
### Description
Added docs on how we can added charmil commands to a host cli. And added base functionality to demonstrate this.

### Type of change

- [x] Documentation change
- [x] Feature


wtrocki:
```
charmilCommands, err := commands.CharmilCommands()
``` 

This part needs to be in the core and return some empty command for now!

jackdelahunt:
> This part needs to be in the core and return some empty command for now!

So have this part of our api but just return a command that does nothing and as not subcommands?

Just wondering because  I am not sure

wtrocki:
Yes.. we need to document something that exist although might not be functional

jackdelahunt:
Okay added the charmil commands in this pr that don't do anything. Also changed how they are added to the root to keep the error handling closer to what go programmers use.

wtrocki:
World class work @jackdelahunt 

Thank you for improving our core framework and applying docs/api first strategy. It is been pleasure.

core/commands/charmil.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/spf13/cobra"
+)
+
+func AttachCharmilCommands(hostRoot *cobra.Command) error {
+
+	if !inDev {
+		return nil
+	}
+
+	cmd := &cobra.Command{
```
wtrocki, line 13:
```suggestion
         // Placeholder command. To be replaced later by actual commands
	cmd := &cobra.Command{
```

PR #108, opened 2021-07-01
feat: add config management package

### LINKED ISSUE #69: Configuration storage and abstraction for plugin

namit-chandwani:
Closes #69 
Based on discussion: https://github.com/aerogear/charmil/discussions/100

### Description
Added a config package that offers a convenient mechanism for both host and plugin developers to manage configurations in their CLIs.

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed


core/config/config.go
```diff
@@ -0,0 +1,91 @@
+package config
+
+import (
+	"fmt"
+
+	"github.com/spf13/viper"
+)
+
+// File defines the fields required to point to a local config file
+type File struct {
```
wtrocki, line 10: Too generic name. Would this be used for plugins?

core/config/config.go
```diff
@@ -0,0 +1,91 @@
...
+	// REQUIRED if the config file does not have the extension in the name
+	Type string
+
+	// Path to look for the config file in
+	Path string
+}
+
+// Handler represents a wrapper around Viper
+type Handler struct {
+	vp *viper.Viper
```
wtrocki, line 25: I'm not sure what viper brings for us?
What is the value?

docs/src/charmil_config.md
```diff
@@ -0,0 +1,170 @@
+# Charmil Config
+
+The Charmil Config package offers a convenient mechanism for both host and plugin developers to manage configurations in their command-line interface (CLI) applications made using [Cobra](https://github.com/spf13/cobra).
+
+## Features:
+
+- Uses [Viper](https://github.com/spf13/viper) under the hood.
+- Helps in maintaining all available configurations in a single central-local config file.
+- Provides the plugin developers with a set of methods to maintain a configuration map and export it to the host CLI.
```
wtrocki, line 9: Wow. Very cool!

docs/src/charmil_config.md
```diff
@@ -0,0 +1,170 @@
...
+- Provides the host CLI developers with a set of methods to import plugin configuration maps and read/write configurations from/to a local config file.
+
+## Steps to Use:
+
+- ### For host CLI developers:
```
wtrocki, line 14: We need to describe architecture.in separate.doc:

- what is host cli
- what is plugin

Separate issue/PR please

docs/src/charmil_config.md
```diff
@@ -0,0 +1,170 @@
...
+      _Example:_
+
+      ```go
+      err := h.Load()
+      if err != nil {
+      	log.Fatal(err)
+      }
+      ```
+
+  6.  Use the `SetValue` and `GetValue` methods respectively to set and get values to/from the current config.
```
wtrocki, line 55: So we cannot have strongly typed config.(as structure). That will be deal breaker for most of CLIs

docs/src/charmil_config.md
```diff
@@ -0,0 +1,170 @@
...
+      ```go
+      h.SetValue("key0", "val0")
+
+      val, err := h.GetValue("key0") // returns "val0"
+      if err != nil {
+      	log.Fatal(err)
+      }
+      ```
+
+  7.  Map a plugin name to its config map using the `SetPluginCfg` method by passing the name of the plugin (as you want it in the config file) as the first argument and the config map imported from the plugin, as the second argument.
```
wtrocki, line 68: Later phase this could be automated by plugin itself

docs/src/charmil_config.md
```diff
@@ -0,0 +1,170 @@
...
+
+      _Example:_
+
+      ```go
+      h.SetPluginCfg("pluginName", pluginCfg)
+      ```
+
+  8.  Call the `MergePluginCfg` method to store the config of every plugin (ie. mapped using the `SetPluginCfg` method) into the current config.
```
wtrocki, line 76: At this point I think our end user api is too low level.
Devs should not merge configs - our code should

docs/src/charmil_config.md
```diff
@@ -0,0 +1,170 @@
...
+
+      _Example:_
+
+      ```go
+      h.MergePluginCfg()
+      ```
+
+  9.  Write current config into the local config file using the `Save` method.
```
wtrocki, line 84: So this is approach where we migrate one time only config..

I do see it too much of hassle.

Lets put all things in the table:

1) some clis hqve their own config and will work with it using structure - like rhoas cli does

2) then cli can become plugin and have the same form of config but saved to the single file

For simplicity we assume that plugin cli was never installed on os and only host is.

What I do not get is - why host needs to interact with plugin and use non typed api (strings).

My understanding was that all we need to do is to provide dynamic config location for fetching config.

So config of plugin can be:

- entire file on it's own
- some subpath of json when running in host

Missing anything?

> Typed by one hand while feeding newborn. Sorry for typos

N

wtrocki: Since plugin is inside host we can also import config structures into host.

docs/src/charmil_config.md
```diff
@@ -0,0 +1,170 @@
...
+      ```
+  3.  Store a new instance of the Charmil Config handler by calling the `New` function.
+
+      _Example:_
+
+      ```go
+      h := config.New()
+      ```
+
+  4.  Use the `SetValue` and `GetValue` methods respectively to set and get values to/from the current config.
```
wtrocki, line 111: Imagine `rhoas kafka topics` commands becoming a plugin. This api will be huge downgrade and require refactor

docs/src/charmil_config.md
```diff
@@ -0,0 +1,170 @@
...
+
+  - _`config.yaml`_ file:
+
+    ```yaml
+    key1: "val1"
+    key2: "val2"
+    key3: "val3"
+    ```
+
+  - Plugin A's Config Map:
```
wtrocki, line 146: Too strong API. If someone wants plugin to be embedded into foobar that will not work. We also have some exisitng config that we could map

wtrocki:
@namit-chandwani I think that I totally missunderstood your proposal from discussion. Sorry for that.

Good news is that I see some useful code that with minor adjustments will work for us.

Bad news is that I feel like you have done much more than that and it might not be needed.

This PR is very promissing so no worries.
Be prepared for +100 comments

core/config/config.go
namit-chandwani: > Would this be used for plugins?

No, since the local config file will be present in the host CLI environment, this will we used over there.

Resolved via commit: https://github.com/aerogear/charmil/pull/108/commits/1eb64335a56861836a41f76cbbc5e6c7becf119f

docs/src/charmil_config.md
wtrocki: Can you reply to this?

namit-chandwani:
Summary:

| - 	|   Issues with my current implementation	|   Reasons	|  Solutions	|   Status	|
|---	|---	|---	|---	|---	|
|  1. 	|   Using maps to store config in both host and plugin	|  Map is untyped and dynamic whereas we need a strongly typed config	|   Use a struct instead	|   -	|
|  2.	|   Storing plugin config in a map and exporting it to host	|  *Same as reason for pt.1* 	| <ul><li>Allow plugins to read/write config values from/into their own set of config structs</li><br><li>Let plugin to be embedded under any key of json</li></ul>|   -	|
|  3.	|   Host CLI developers need to merge plugin configs into the central config file	|   Unnecessary extra work for host CLI developers, which can be easily avoided by some changes in code	|   *Solution for pt.2 will more or less solve this issue too*	|   -	|

docs/src/charmil_config.md
namit-chandwani: Okay so I revisited the Rhoas config implementation and I think now I understand what the issues with my current approach are. Will change all the things that you've mentioned.

I've tried to summarize all the current issues with reasons and solutions for the same, through this comment: https://github.com/aerogear/charmil/pull/108#issuecomment-873180768

Please have a look and let me know if there is anything to be added/removed :)

wtrocki:
Needs rebase? Is that ready for review?

core/config/config.go
```diff
@@ -0,0 +1,164 @@
...
+	"gopkg.in/yaml.v2"
+)
+
+type Handler struct {
+	cfg      interface{}
+	filePath string
+	fileExt  string
+}
+
+func New(path string, cfg interface{}) *Handler {
```
jackdelahunt, line 20:
```suggestion
func NewHandler(path string, cfg interface{}) *Handler {
```
Generally constructors in go are called NewThing(...). This is nit-picky I know.

core/config/config.go
```diff
@@ -0,0 +1,164 @@
...
+
+	err = h.unmarshal(buf, h.cfg)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (h *Handler) Save() error {
```
jackdelahunt, line 56: In it's current form (maybe you have plans to change it) but this Save function overwrites all data in the config file right? Wouldn't we just want to save the values defined in the host's config struct and leave the plugin data.

examples/host/main.go
```diff
@@ -3,23 +3,75 @@ package main
...
+
+	// Loads config values from the local config file
+	err := h.Load()
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+func main() {
+	// TODO: Initialize the factory struct here and pass it to the plugins
```
jackdelahunt, line 49: So is the idea the plugin will be able to load saved data from the config file when this gets set up.

wtrocki:
> So is the idea the plugin will be able to load saved data from the config file when this gets set up? And also save?

Yes. Both read and save should work with the plugin host scenarios

jackdelahunt:
Also make sure to update the docs before next review.

namit-chandwani:
> Needs rebase. Is that ready for review?

Just making some last-minute changes and updating docs right now. Will rebase and ping you once done

jackdelahunt:
> Yes, doing that right now :)

Thanks, make sure to take your time and enjoy it. No need to put a lot of pressure on yourself just because we want to review.

core/config/config.go
namit-chandwani: I had used it in the POC initially, just for simplicity.

Its no longer a part of my current updated implementation.

namit-chandwani:
Forgot to ping you earlier 😅

This is ready for review and the changes discussed in review comments have been made.

CC @wtrocki @jackdelahunt

examples/host/main.go
namit-chandwani: Yes, you're absolutely right.

Just that we won't need to pass the `Handler` as a whole into the factory. Instead, only including the local config file path in the factory and passing it to the plugins will suffice. WDYT?

docs/src/charmil_config.md
```diff
@@ -0,0 +1,180 @@
...
+      cfg.Key8 = "val8"
+
+      // Overwrites a value in config
+      cfg.Key7 = "val7"
+
+      // Returns the value under specified key in config
+      fmt.Println(cfg.Key6) // Prints: val6
+      ```
+
+  5.  Use the `MergePluginCfg` function to merge the current plugin config into the local config file
```
wtrocki, line 124: So far all was really good, but I feel that this merge can be done by including structure of the plugin config inside host config. This way we get serialization sorted out of the box. (Requirement: it needs to be json is fine for now)

wtrocki (commented):
Very good step forward. I left some questions about merging config. Everything else looks ok. 
I need to play with this myself more once we merge it

examples/host/main.go
jackdelahunt: Yeah I think your right, just storing the path would make more sense to me.

jackdelahunt (commented):
From what I can see besides the points that have already been raised this is looking really good so far. Good work!

namit-chandwani:
TODO:
- [ ] Rename config handler struct
- [ ] Modify the `MergePluginCfg` function
- [ ] Modify the `Save` method
- [ ] Update docs

core/config/config.go
```diff
@@ -0,0 +1,213 @@
...
+	"path/filepath"
+
+	"github.com/imdario/mergo"
+	"github.com/pelletier/go-toml"
+	"github.com/tidwall/sjson"
+	"gopkg.in/yaml.v2"
+)
+
+// Handler defines the fields required to manage config.
+type Handler struct {
```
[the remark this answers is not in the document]
namit-chandwani: Resolved via commit: https://github.com/aerogear/charmil/pull/108/commits/822770b48fb09df32d9abd63f1a6866520eb9219

docs/src/charmil_config.md
namit-chandwani: Resolved via commit: https://github.com/aerogear/charmil/pull/108/commits/a33727536fea8cbaafeae25b96ce55eaab12ef72

wtrocki (approved):
I really hate to be this guy comming back and forth on this. I trust you have applied all suggestions. Lets merge it and refine it on main branch

wtrocki:
Move todos to issues so build will pass

core/config/config.go
```diff
@@ -0,0 +1,213 @@
...
+
+	return nil
+}
+
+// Save writes config values from the linked instance
+// of host CLI config struct to the local config file
+// (using the file path linked to the handler).
+func (h *Handler) Save() error {
+	// To store the current contents of the local config file
+	dst := &map[string]interface{}{}
```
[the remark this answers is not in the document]
namit-chandwani: Resolved via commit: https://github.com/aerogear/charmil/pull/108/commits/7f08531217e0b98adbfdcc00f06147d6879be1b8

namit-chandwani:
@wtrocki This PR is ready to merge now.
I've made all the required changes and removed the TODO comments (will create new issues for them shortly), so that build could pass.

wtrocki:
I do not want to see it anymore 🙈

examples/host/main.go
namit-chandwani: @jackdelahunt BTW just to update you, according to the latest implementation, we will be including the entire `Handler` in the factory (just as you had mentioned earlier).

Will create an issue/PR for the same shortly.

PR #127, opened 2021-07-03
Feature: validator command for use with dev builds

jackdelahunt:
### Description
Added a command called validate that can be used when a host build with the dev tag. This command runs validate on all non-charmil commands that the host command contains.

### Type of change

### Checklist

- [ ] Code Review completed

## Screenshots
Validate command is ran on a dev build which prints the commands added by the host that fail validation based on the default rule config.
image

### DIFF
--- a/core/commands/charmil.go
+++ b/core/commands/charmil.go
@@ -10,12 +10,19 @@ func AttachCharmilCommands(hostRoot *cobra.Command) error {
 		return nil
 	}
 
-         // Placeholder command. To be replaced later by actual commands
+	// Placeholder command. To be replaced later by actual commands
 	cmd := &cobra.Command{
 		Use:     "charmil",
 		Short:   "built in charmil commands",
 		SilenceUsage: false,
 	}
+
+	// charmil custom commands
+	validateCommand ,err := ValidateCommand()
+	if err != nil {
+		return err
+	}
+	cmd.AddCommand(validateCommand)
 
 	hostRoot.AddCommand(cmd)
 	return nil
--- a/core/commands/validator.go
+++ b/core/commands/validator.go
@@ -1,0 +1,33 @@
+package commands
+
+import (
+	factory2 "github.com/aerogear/charmil/core/factory"
+	"github.com/aerogear/charmil/validator/rules"
+	"github.com/spf13/cobra"
+)
+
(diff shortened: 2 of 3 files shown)


jackdelahunt:
Looking for suggestions on this pr. Explicetly checking for charmil name [here](https://github.com/aerogear/charmil/pull/127/files#diff-c5c4f57f7d74bd81639f9c9ac756d9f399dc576af3eb65d02bfefa68fb2fb906R55-R57) is pretty messy plus you have to do it when running the host's children. This could be another +1 for making charmil commands their own root.

wtrocki:
Charmil is 

- Plugin system that enables developers to build complex multi repository clis

- Validator that runs during developent time that ensures that cli meets criteria set by host cli etc.

- Distribution/Registry of clis that can be dynamically embeded (This is currently suspended until we get first two)


TLDR?
Componetize and reuse
Validate
Distrubute

We need that info in readme.

Now.. this PR doesnt fit any of those use cases as it add code to cli that will be included in end user CLI..They do not need to validate it.

Idea is also that plugins validate that during dev time so host do not need it. As mentioned before I'm not sure if we go this way it will be little bit confusing for devs and in our docs

core/commands/charmil.go
```diff
@@ -10,13 +10,20 @@ func AttachCharmilCommands(hostRoot *cobra.Command) error {
...
 
-         // Placeholder command. To be replaced later by actual commands
+	// Placeholder command. To be replaced later by actual commands
 	cmd := &cobra.Command{
 		Use:     "charmil",
 		Short:   "built in charmil commands",
 		SilenceUsage: false,
 	}
 
+	// charmil custom commands
```
wtrocki, line 20: Confusing - Charmil commands will be dedicated to end user.
Like for registry it will be view plugins to install or manage config.
Validator is not charmill SDK - it is separate component

wtrocki:
> This could be another +1 for making charmil commands their own root.

That was my assumption but only for validator.

we have 2 types of commands:

- User facing 
- Development/tooling

We should not mix them as it will be uber confusing for anyone to wrap head around it.

I think this PR can be adapted easily so no worries

wtrocki:
This brought good discusion. I think we can close this for now

PR #139, opened 2021-07-05
feat: useMatches rule for Use

### LINKED ISSUE #84: Regex validation for command name
useMatches that accept valid regexp that checks command name
Provide rule that validates if cmd doesn;t have _ or - etc. 

related to #58

ankithans:
Closes #84 

### Description
- UseMatches rule for validating use with default or user provided regex

### Type of change
 
- [x] Documentation change

### Checklist

- [ ] Code Review completed

### DIFF
--- a/validator/rules/config.go
+++ b/validator/rules/config.go
@@ -23,8 +23,9 @@ type ValidatorOptions struct {
 // ValidatorRules consists of all the rules
 // present in validator
 type ValidatorRules struct {
-	Length    `json:"Length"`
-	MustExist `json:"MustExist"`
+	Length     `json:"Length"`
+	MustExist  `json:"MustExist"`
+	UseMatches `json:"UseMatches"`
 }
 
 // ExecuteRules executes all the rules
@@ -58,6 +59,9 @@ func ValidatorConfigToRuleConfig(validatorConfig *ValidatorConfig, ruleConfig *R
 				Verbose: defaultVerbose,
 				Fields:  map[string]bool{"Use": true, "Short": true, "Long": true, "Example": true},
 			},
+			UseMatches: UseMatches{
+				Regexp: `^[^-_]+$`,
+			},
 		},
 	}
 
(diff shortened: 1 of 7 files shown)


validator/rules/use_matches.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// Validate is a method of type Rule Interface
+// which returns validation errors
+func (u *UseMatches) Validate(cmd *cobra.Command) []validator.ValidationError {
+	return u.validateUse(cmd)
+}
+
+// validateUse compares the regexp with Use attribute
+func (u *UseMatches) validateUse(cmd *cobra.Command) []validator.ValidationError {
```
jackdelahunt, line 31: If we have more then one thing to do in our `Validate` function it is good to pass them off to new ones but this `validateUse` just seems unnecessary for now.

ankithans: let me change that in all rules, if required?

jackdelahunt:
It would also be nice to see this working. I know you have probably tested this yourself but maybe adding a test inside of the validator directory just to get a feel for what it looks like in practice.

validator/rules/use_matches.go
jackdelahunt: Sure, I can see in the `lengthRule` and `mustExistRule` you have so I think it would be best to quickly change them.

ankithans: done

ankithans:
@jackdelahunt are you talking about unit tests or example here https://github.com/aerogear/charmil/blob/regex-rule/validator/example/cmd_test.go#L14

jackdelahunt:
> Are you talking about unit tests or example here

Yes, maybe just add another unit test below it.

validator/rules/config.go
```diff
@@ -58,6 +59,9 @@ func ValidatorConfigToRuleConfig(validatorConfig *ValidatorConfig, ruleConfig *R
 				Verbose: defaultVerbose,
 				Fields:  map[string]bool{"Use": true, "Short": true, "Long": true, "Example": true},
 			},
+			UseMatches: UseMatches{
+				Regexp: `^[^-_]+$`,
```
wtrocki, line 63: Some rules might not need defaults. Is that default you would like to enforce?

ankithans: yes this is default, we can find some good default value, or can leave it empty

jackdelahunt: Maybe a default that checks for line breaks or EOF. Something basic that everyone wouldn't want.

ankithans: kept if empty for now!

wtrocki:
+100 for empty..Amazing work guys

PR #144, opened 2021-07-06
feat: add ignore command option

### LINKED ISSUE #86: Config with array of commands during validation
provide config with array of commands to skip/ignore validation

ankithans:
Closes #86 

### Description
IgnoreCommands map[string]bool option in ValidatorOptions, for ignoring validation

### Type of change

- [x] Documentation change

### Checklist

- [ ] Code Review completed

### DIFF
--- a/validator/rules/executor.go
+++ b/validator/rules/executor.go
@@ -26,6 +26,13 @@ func ExecuteRulesInternal(cmd *cobra.Command, ruleConfig *RuleConfig, userValida
 	var errors []validator.ValidationError
 	info := validator.StatusLog{TotalTested: 0, TotalErrors: 0, Errors: errors}
 
+	// if command needs to be ignored
+	if val, ok := userValidatorConfig.ValidatorOptions.SkipChildren[cmd.CommandPath()]; ok {
+		if val {
+			return info.Errors
+		}
+	}
+
 	// initialize default rules
 	initDefaultRules(userValidatorConfig, ruleConfig)
 
@@ -36,6 +43,7 @@ func ExecuteRulesInternal(cmd *cobra.Command, ruleConfig *RuleConfig, userValida
 }
 
 func executeHelper(cmd *cobra.Command, info *validator.StatusLog, ruleConfig *RuleConfig, userValidatorConfig *ValidatorConfig) []validator.ValidationError {
+
 	info.Errors = executeRecursive(cmd, info, ruleConfig, userValidatorConfig)
 
 	// prints additional info for the checks
@@ -77,7 +85,7 @@ func executeRulesChildren(cmd *cobra.Command, info *validator.StatusLog, ruleCon
 func validate(cmd *cobra.Command, info *validator.StatusLog, ruleConfig *RuleConfig, userValidatorConfig *ValidatorConfig) {
 
 	// if command needs to be ignored
-	if val, ok := userValidatorConfig.ValidatorOptions.IgnoreCommands[cmd.Use]; ok {
+	if val, ok := userValidatorConfig.ValidatorOptions.SkipCommands[cmd.CommandPath()]; ok {
 		if val {
 			return
 		}
--- a/validator/rules/executor.go
+++ b/validator/rules/executor.go
@@ -30,13 +30,13 @@ func ExecuteRulesInternal(cmd *cobra.Command, ruleConfig *RuleConfig, userValida
 	initDefaultRules(userValidatorConfig, ruleConfig)
 
 	// validate the root command
-	validate(cmd, &info, ruleConfig)
+	validate(cmd, &info, ruleConfig, userValidatorConfig)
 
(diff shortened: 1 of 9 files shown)


ankithans:
@aerogear/charmil-core

validator/rules/executor.go
```diff
@@ -46,35 +46,42 @@ func executeHelper(cmd *cobra.Command, info *validator.StatusLog, ruleConfig *Ru
...
 	}
 	return info.Errors
 }
 
 // validate returns validation errors by executing the rules
-func validate(cmd *cobra.Command, info *validator.StatusLog, ruleConfig *RuleConfig) {
+func validate(cmd *cobra.Command, info *validator.StatusLog, ruleConfig *RuleConfig, userValidatorConfig *ValidatorConfig) {
+
+	// if command needs to be ignored
+	if val, ok := userValidatorConfig.ValidatorOptions.IgnoreCommands[cmd.Use]; ok {
```
wtrocki, line 80: How this will work with nested commands? Do we use `.` to map the nested commands.

Example:
```
kafka.create 
registry.create
```

ankithans: it captures `Use` of the command, so 
```
kafka.create 
registry.create
```
all `create` subcommands will be ignored. Let's fix this!

wtrocki: However would that be hard. Can we simply create method that returns true or false for the current command and rule rule?

ankithans: for nested commands, best way is to use CommandPath() provided by cobra. 
user will specify : `{"kafka create": true, "echo": true}`

This config is coming to my mind atm -
```go
type ValidatorOptions struct {
	Verbose      bool            
	SkipChildren map[string]bool 
	SkipCommands map[string]bool 
}
```
RuleOptions to be present in each rule, for controlling rule specific properties/options
```go
type RuleOptions struct {
	Verbose      bool
	SkipCommands map[string]bool
}
```
eg - 
```go
type Length struct {
	RuleOptions validator.RuleOptions
	Limits      map[string]Limit `json:"Limits"`
}

Comment (wtrocki) [reply validator/rules/executor.go:80]: type Length struct {
	RuleOptions validator.RuleOptions
	
Could that be part of the interface so we can still have generic processing?

Comment (ankithans) [reply validator/rules/executor.go:80]: > RuleOptions validator.RuleOptions
Could that be part of the interface so we can still have generic processing and enforce options for each rule

Rules interface?

> SkipChildren map[string]bool 
Not needed/confusing. If you specify command path that is grouping other commands all grouped ones will be disabled.

can you please provide an example for understanding

> map[string]bool 
> Not sure if that would give us enough to control command (path to command), rule, level - Error/Ignore (could be warn in the future)

do we need all this details, when we ignore some command?

Comment (wtrocki) [reply validator/rules/executor.go:80]: > do we need all this details, when we ignore some command?

Agreed. No need

**Comment by wtrocki** [comment]
I'm good to merge this :)

**Comment by ankithans** [comment]
>I'm good to merge this :) after we add documentation

Yes, but shouldn't these be included as well?
```
Ignore entire chain of commands (when you do dev etc.) and still want tests to pass
Ignore single command in nested scenario for all rules
Ignore single command in nested scenario for some rule.
```

**Comment by wtrocki** [comment]
You can try adding some quick wins :D

**Comment by ankithans** [comment]
> Working

## Ignore Commands
Sometimes during development, you want to pass the tests for certain commands, but at the same time use Validator for tests. Validation can be skipped/ignored for the commands, mentioned in the validator configuration.
To ignore the commands you need to specify the path of the command in validator configuration.

1. Skip single command
```go
ValidatorOptions: rules.ValidatorOptions{
	SkipCommands: map[string]bool{"cmd100 cmd0 subcmd01": true},
},
```

2. Skip the command including all children
```go
ValidatorOptions: rules.ValidatorOptions{
	SkipChildren: map[string]bool{"cmd100": true},
},
```

3. Skip the command for specific rule
```go
Length: rules.Length{
	RuleOptions: validator.RuleOptions{
		SkipCommands: map[string]bool{"cmd100": true},
	},
	Limits: map[string]rules.Limit{
		"Use": {Min: 1},
	},
},

File: website/docs/validator.md [hunk 2]
```diff
@@ -42,4 +42,33 @@ for _, errs := range validationErr {
...
+ValidatorOptions: rules.ValidatorOptions{
+	SkipChildren: map[string]bool{"mycli": true},
+},
+```
+
+3. Skip the command for specific rule
+```go
+Length: rules.Length{
+	RuleOptions: validator.RuleOptions{
+		SkipCommands: map[string]bool{"mycli actions create": true},
```
wtrocki, line 69: @ankithans is that ok?

ankithans: yes.. this looks good!

wtrocki: example had only `create` command so I wasn't sure if you are using cobra paths for this

Issue #143, opened 2021-07-06
Generate boilerplate starter with charmil commands


ankithans:
As we will provide some starter examples to start developing, we may provide a CLI tool to generate a starter project/plugin instead of copying, just like cobra does.

wtrocki:
Good idea. I'm working on pulling out the bits for initial project structure. There are common elements that can be templated like:

- cmd name
- repository name
etc.

https://github.com/Apicurio/apicurio-cli

Will provide more info about this once I got something working soon. Full blown generation might be too much but we can figure out something that will allow us to swap values

ankithans:
Remove unnecessary feats from core and put them in starter. and Set up starter with Charmil core features
- Localization
- Autocompletion commands

PR #145, opened 2021-07-06
chore: starter template for the cli without charmil

wtrocki:
This PR introduces starter plugin CLI codebase.
Codebase contains number of elements duplicated with core charmil - factory that we can move to charmil SDK once all those concepts are there. 

Creating for early feedback. This is not going to get merged due to number of problems:
- go submodule is not working
- cli name is hardcoded and not easy to change
- root command is hardcoded 
- Multiple elements in the PR needs to be pulled to SDK.
- This clashes with current examples
- Missing documentation and usage docs - requires more work - See #143 

 ## Verification

```
cd starter
make install
starter 
```


wtrocki:
Idea - Starter might be separate repository (this will require us to do frequent releases so IMHO we can merge that now into charmil and get it moved if needed.

wtrocki:
We might change cli name to be something like smpl or other non english world for simple search replace.

wtrocki:
Looking for someone picking up this to integrate with charmil commands

ankithans:
Like `charmil init CLI_NAME` can create a full fledged starter project, using predefined starter template
I want to take it over. want to work with go templates from a long time

wtrocki:
but for that we need yet another thing: charmil cli to manage CLI to build CLIs :)

ankithans:
okay, user can run `make setup` after cloning etc.. CLI can make this more easier just go get charmil and init. We can go for any as per suggestions.
I am creating issues, which needs to be done before this

jackdelahunt:
Looks nice and is a good place to start. Just have a few questions/comments.

- Moving to a separate repo and changing name make sense with this for reasons said before and cloning the repo of the SDK to get the SDK seems weird
- Using a make command along the lines of `make charmil/init NAME=smpl` in this smpl repo sounds like a really nice feature as I don't think a separate CLI is necessary here
- @wtrocki can you just explain what do you mean by integrating this with charmil commands as I can't see what this looks like right now
- Creating a separate repo for this and start to add charmil to it as we go/add features makes sense instead of doing that all at once

ankithans:
> @wtrocki can you just explain what do you mean by integrating this with charmil commands as I can't see what this looks like right now

https://github.com/spf13/cobra/tree/master/cobra
i think have a look at cobra cmd tool. It has utilities such as cobra init, cobra add cmd, etc. which comes with cobra's recommended settings with viper and all. Here we do have similar thing, we want the user to follow the charmil SDK and for easing this process, CLI tool is better(will rm the need for separate repo). But makefile will do the similar thing

wtrocki:
> https://github.com/spf13/cobra/tree/master/cobra

Actually good point. Since I'm going to PTO soon let's have final call what we do with this:

1. Do we need separate repository for this now and move this source code?

2. Do we merge that without review and refine it on main later on to swap with charmil SDK?

jackdelahunt:
> https://github.com/spf13/cobra/tree/master/cobra

Ah okay I thought they would have to clone our whole repo but this looks really cool and would be a really nice feature for poeple to use.

wtrocki:
Cool. Enjoy playing with this!

PR #153, opened 2021-07-07
refactor: include config handler in factory

namit-chandwani:
### Description
Based on a TODO comment dropped while working on the config package (Ref: [Link](https://github.com/aerogear/charmil/pull/108/commits/4a2f025b6431a88fa4b1ad31d1bece7ac11733eb#diff-13338daeaca64c71d39b900e5c4e96bb600794eebe6df6588a607f051970aaecL55))

Changes made:
- The `Factory` struct should be initialized in the host CLI and passed on to the plugins. But in the current setup of host and plugin examples, the `Factory` was being initialized in the plugins themselves. Added a fix for that.
- Included the `CfgHandler` struct in `Factory` (**as discussed here**: https://github.com/aerogear/charmil/pull/108#discussion_r664903307)
- Modified the `echo` example plugin to use the Config package
- Replace dummy config values in the `adder` example plugin with actual configurations

### Type of change

- [ ] Documentation change
- [x] Other (please specify): Refactor

### Checklist

- [ ] Code Review completed

### DIFF
--- a/examples/plugins/adder/main.go
+++ b/examples/plugins/adder/main.go
@@ -23,7 +23,7 @@ var cfg = &config{}
 
 // AdderCommand returns the root command of plugin.
 // This will be added to the host CLI as an extension.
-func AdderCommand(h *c.CfgHandler, f *factory.Factory) (*cobra.Command, error) {
+func AdderCommand(f *factory.Factory) (*cobra.Command, error) {
 
 	// Stores the config for localizer
 	cfg.LocConfig = localize.Config{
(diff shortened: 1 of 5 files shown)


examples/plugins/adder/main.go
```diff
@@ -15,45 +15,40 @@ import (
...
 		return nil, err
 	}
 
-	//Stores a new factory instance with default settings
-	f := factory.Default(loc)
+	opts := &factory.Factory{
+		Logger:     f.Logger,
+		Localizer:  loc,
+		CfgHandler: f.CfgHandler,
+	}
```
jackdelahunt, line 45: I can see why we do this. We need the logger and config from the host factory but we need to setup a localizer based on the plugin. I just find this weird looking but honestly not sure what else to do besides this either.

jackdelahunt (commented):
Besides the comment I left about creating the new factory which I am not sure to really with either; everything looks fine here and works as I imagine this should.

examples/plugins/adder/main.go
namit-chandwani: Agreed.
This is just a temporary workaround to get things going for the time being. But we need to think of a better way to handle this in future. 

I'll create a separate issue for discussion on this, tomorrow.

Issue #149, opened 2021-07-07
Pull pkg from starter to charmil core


ankithans:
#145 gives us a starter project, in which many elements should be present in charmil core.

wtrocki:
It should be other way around (I think you meant that) - we should be removing things from starter and replace them with core SDK.

I already created issues for missing elements that are not on starter but can be in the core SDK and later integrated into starter

https://github.com/aerogear/charmil/issues/138
https://github.com/aerogear/charmil/issues/136
https://github.com/aerogear/charmil/issues/135
and couple more

ankithans:
> we should be removing things from starter and replace them with core SDK.

yep, we can do them one by one, this can be our current priority now

PR #154, opened 2021-07-08
Refactor core

ankithans:
### Description
Extract packages from starter and put them into charmil-core
Next Step would be to make a starter CLI able to run all these packages!
after that a starter template -> then CLI tool

I will provide documentation for all the packages, once i get them running in the example


wtrocki:
@ankithans Starter have opiniated libraries from rhoas bringing them to core might be challenging.
It is way easier to figure out core/config etc. and swap it with what is there

https://github.com/aerogear/charmil/issues/149

wtrocki:
I thino this PR has tons of good changed:

- Removes elements that are already in core
- Swaps some elements with charmil

I will leave others for now as this PR might be too large to deal with all at the same time.

wtrocki (approved):
Good to go after cleanup and rebase. 
World class job

wtrocki:
@namit-chandwani we are missing only config

ankithans:
> Removes elements that are already in core
Swaps some elements with charmil

yes it does these changes and next I will make these new features in core - running with starter example

wtrocki:
I will keep core small, well tested, well documented. PRs to core should be small and targeted.

wtrocki:
Good to go!

ankithans:
so are you recommending to implement/test these additions to core in this PR? we can do that as well!

wtrocki:
> so are you recommending to implement/test these additions to core in this PR? we can do that as well!
This may be better as we will not have any unused code in main. All things will be well tested and documented before adding to main.

No I recommend to minimize amount of untested code landing into core without issues/discussion. 
Example: Colours - I do not see it being particularly useful but that is just my opinion.

Issue #151, opened 2021-07-07
SkipChildren option not working in validator


ankithans:
SkipChildren only works for root command, not for command path. 
For eg. root command for rhoas-cli is `rhoas`. It works for this, but not for `rhoas kafka` (contains ..list, ..create etc)

wtrocki:
We need some basic unit tests to see this type of issues before we release.

ankithans:
actually right, this happened for more than 2 times, I can do this 👍

wtrocki:
This is also duplicate of removing examples. We rely on examples to test things. My take is that unit tests can have some very bare commands structures etc. thus giving better testing capabilities.

Issue #136, opened 2021-07-05
Extract documentation helpers into charmil core


wtrocki:
### Feature or problem description
 

See: https://github.com/redhat-developer/app-services-cli/blob/main/pkg/doc/adoc_docs.go#L15

ankithans:
use cobra doc generator in starter

Issue #135, opened 2021-07-05
[Generic flags] Debug


wtrocki:
### Feature or problem description

See
https://github.com/redhat-developer/app-services-cli/blob/main/pkg/cmd/debug/flag.go#L3

Issue #134, opened 2021-07-05
[Generic commands] Support for completion


wtrocki:
Copy implementation from: 
 https://github.com/redhat-developer/app-services-cli/blob/main/pkg/cmd/completion/completion.go#L14:L14

ankithans:
need to provide as part of starter, not from core

PR #155, opened 2021-07-09
charmil CLI for starter project

### LINKED ISSUE #143: Generate boilerplate starter with charmil commands

ankithans:
Closes #143 

> WIP may be not ready for review yet

### Description

Charmil CLI for creating starter Charmil project

### Type of change

- [x] Documentation change

### Checklist

- [x] starter example for CLI
- [x] include charmil validator tests in starter
- [x] starter go template for CLI files
- [x] charmil CLI basic structure
- [ ] Adding license, contibuting.md and other no golang stuff
- [ ] charmil CLI init command
- [ ] Code Review completed


ankithans:
@aerogear/charmil-core if someone wants to comment on this https://github.com/aerogear/charmil/tree/starter/starter-1 structure, like making it better(for now it is without config, we can add it after we figure it out)

starter-1/cmd/abc/main.go
```diff
@@ -0,0 +1,52 @@
...
+	buildVersion := build.Version
+	// init factory
+	cmdFactory := factory.Default(localizer)
+
+	// root command
+	rootCmd := root.NewRootCommand(cmdFactory, buildVersion)
+	rootCmd.InitDefaultHelpCmd()
+
+	return rootCmd, cmdFactory
+}
```
jackdelahunt, line 42: I think using a `init()` function here that populates values defined in var would look a bit cleaner. Namit does this in his config pr

starter-1/pkg/cmd/version/version.go
```diff
@@ -0,0 +1,43 @@
...
+		},
+	}
+
+	return cmd
+}
+
+func runCmd(opts *Options) (err error) {
+	fmt.Fprintln(opts.IO.Out, opts.localizer.LocalizeByID("version.cmd.outputText", localize.NewEntry("Version", build.Version)))
+	return nil
+}
```
jackdelahunt, line 43: Outsourcing the call back here is much cleaner then before aswell

jackdelahunt:
My last comment about the `starter-1` CLI is I am a bit confused on why do:
cmd -> abc -> ...
pkg -> cmd -> ...

I think removing the cmd directory around abc is a bit cleaner

starter-1/cmd/abc/main.go
ankithans: sure

ankithans:
> My last comment about the starter-1 CLI is I am a bit confused on why do:
cmd -> abc -> ...
pkg -> cmd -> ...

basic golang project struct, followed by https://github.com/redhat-developer/app-services-cli/ https://github.com/cli/cli and many others.

starter-1/pkg/cmd/root/root.go
```diff
@@ -0,0 +1,26 @@
+package root
+
+import (
+	"github.com/aerogear/charmil/core/factory"
+	"github.com/spf13/cobra"
+
+	Cliversion "github.com/aerogear/charmil/starter-1/pkg/cmd/version"
```
namit-chandwani, line 7: `Nitpick`: Maybe we can convert this alias to lowercase. Almost every project that I've come across follows this naming convention. 
```suggestion
	cliversion "github.com/aerogear/charmil/starter-1/pkg/cmd/version"
```


WDYT?

core/build/build.go
```diff
@@ -0,0 +1,94 @@
...
+// Define public variables here which you wish to be configurable at build time
+var (
+	// Version is dynamically set by the toolchain or overridden by the Makefile.
+	Version = "dev"
+
+	// Language used, can be overridden by Makefile or CI
+	Language = "en"
+
+	// RepositoryOwner is the remote GitHub organization for the releases
+	RepositoryOwner = "redhat-developer"
```
wtrocki, line 22: Please do not copy code to core

ankithans: yes, this has to be configurable by CLI. I already made a note for it.. will not push in main with this

starter-1/cmd/abc/main.go
```diff
@@ -0,0 +1,52 @@
...
+
+func main() {
+	cmd, cmdFactory := abc()
+
+	if err := cmd.Execute(); err != nil {
+		fmt.Fprintln(os.Stderr, err)
```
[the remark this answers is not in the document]
ankithans: yep

starter-1/pkg/cmd/root/root.go
ankithans: sure

wtrocki:
@ankithans While I have commented on issue that this is good idea I do not think we should do that before we have core there. Typically I would not pick issues that you have created yourself without dicussion.

We can spend a lots of mentors time reviewing those changes. We need core and starter first.

wtrocki:
Building CLI will take significant amount of time for us. 
There is no sense investing in scafolder and there are existing projects that do scafold starters etc. and we have nothing to scafold right now.

ankithans:
> We need core and starter first.

Yep core demands config and some issues #136 #135 #134. And starter is just ready(but without these mentioned feats, which are necessary) So I proceeded with this CLI tool(dont't think will take much of the time).

>There is no sense investing in scafolder and there are existing projects that do scafold starters etc. and we have nothing to scafold right now.

So ig let's continue this CLI tool once we solve above issues in the core and implement them in starter project!

namit-chandwani:
> Yep core demands config and some issues #136 #135 #134.

I'll try to create a PR for config by today EOD

wtrocki:
Core overal will need full blown refactor and integration with starter. I can help with some suggestions how we can split work.

namit-chandwani:
> I can help with some suggestions how we can split work.

Yes, that would be great

wtrocki:
> Yes, that would be great

Config

@ankithans 

- Validator 
- i18n
- cleanup starter

ankithans:
#### core has 
- build (should be served from factory, need to have CLI specific options like repo details)
- color (can be used with logger, I will try to figure out the best way)
- commands (not needed in core, but in pkg)
- config (not integrated, provided by factory)
- debug (should be served from factory?? to add debug flag to command flags)
- factory (most of the features are provided from here)
- iostreams, localize, logging(works well through factory)

#### need to add
- Documentation helpers
- support for completion 

#### Validator 
- #151
-  #85 
-  #116

what do we need in i18n?

wtrocki:
> color (can be used with logger, I will try to figure out the best way)

I missed us adding that

namit-chandwani:
> * iostreams, localize, logging(works well through factory)

I think we still need to find a way for the localizer to work perfectly when the factory struct is initialized in Host and passed down to Plugins

wtrocki:
We very close to make main a set of unrelated and unrefined code. If we do this then there is no way back.
Cost of cleanup and discussion about "why this code is here" can take all our progress.

To explain things in normal way. We need foundations based on the project and then apply this foundations to some real case (starter - Apicurio + RHOAS)

Imagine that we build home, have bunch of materials laying around and start with pool and playground.

wtrocki:
> I think factory should be specific to the CLI, if not how will the plugin developer get the factory object when there is no host 

Factory is esentiall element of charmil core. It mounts all features and allows to share them between plugin and host. Factory is needed to map specifics of being part of the bigger cli

jackdelahunt:
> It mounts all features and allows to share them between plugin and host

Host passes the factory to the plugin currently. Do we expect the plugin to ever alter the state of factory. Just curious really.

wtrocki:
> Also locales of plugins will be specific to plugins, we cannot provide them from host.

Lets create issue to check this. Not sure if current code will work when some clis are library.
Effectively fix could be to have two internationalizations and keep i18n out of factory etc.

namit-chandwani:
> Do we expect the plugin to ever alter the state of factory. Just curious really.

I'm having the same question. 
Like weren't we suppose to enforce the factory elements on plugins?

namit-chandwani:
> > I think we still need to find a way for the localizer to work perfectly when the factory struct is initialized in Host and passed down to Plugins. (Ref: #153
> 
> Passing localizer back to plugin will not work as plugin will have different folder for i18n lookup

Yes, that was just done in order to get the code running temporarily

> Effectively fix could be to have two internationalizations and keep i18n out of factory etc. 

Exactly, this is what I had in mind too

ankithans:
> i18n could be just in starter and we might not need it in core.

may be provide it from core, but not from factory. This is possible to do.
we can provide internal code for i18n in starter, that will be unnecessary boilerplate, when we can do above!

wtrocki:
> Do we expect the plugin to ever alter the state of factory. Just curious really.

This factory was something we put initially to add features in the way we can consume them. 
Since I have tried this with apicurio I found numerous problems with config, internationalization, logging etc.

To answer this question: Factory is immutable right now. We need to split our work towards some specific problems like config reusability that will require some factory like feature.

Second element wil be to implement helpers - replication of common flags across the cli etc.

This can land into starter and future plugins without factory at all.

wtrocki:
> Yep I was doing this with what we have, and adding other feats later in templates and all, but that is bad idea.

No bad ideas. Sometimes you are really fast with implementation and I need to catch up to understand what this PR does intent to do. Do you mind explaining in the issue what we are trying to do for scafollder

wtrocki:
To close that entire thread in wrong pr: i would rather have i18n + config + validator well done and in starter than bunch of opiniated and unrefined helpers.

@ankithans I see you are passionate about scafolding. Lets refine core and starter with very minimum and then we can work on it for #67 etc

wtrocki:
my take is to keep starter with all of those extra features. We can always pull them to core if needed.
Trick is sdk easy and resolve specific problems.

I18n in sdk doesnt do anything than just wraps it

ankithans:
### Next Steps

- Remove unnecessary feats from core and put them in starter and set up starter with Charmil core features
    - update docs
    - Factory
    - Config
    - Localization
    - Autocompletion commands
    - Validation tests
- Prepare templates for the CLI including no golang stuffs like license etc.
- Generate Starter functionality
- Integrate above with charmil init command

this can be different PR's but there will be nothing which will give any value. So preferring to finish above in this PR!

ankithans:
this PR does
- initialize charmil CLI
- integrates core feats into starter
- include validator tests in starter
- autocompletion commands into starter

starter/go.mod
```diff
@@ -1,12 +0,0 @@
...
-go 1.16
-
-require (
-	github.com/aerogear/charmil v0.2.0
-	github.com/spf13/cobra v1.2.1
-	github.com/spf13/pflag v1.0.5
-	golang.org/x/text v0.3.6
-)
-
-replace github.com/aerogear/charmil => ../../charmil
```
namit-chandwani, line 12: As the starter will be using some released version of the core, so we'll need it to be a sub-module. Which is why `go.mod` will be required here

ankithans: does it matter? as this will be most likely wiped and transformed into templates

Makefile
```diff
@@ -16,7 +16,6 @@ run:
 	go run ./examples/host
 
 build:
-	go build -o host_example ./examples/host
```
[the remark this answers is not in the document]
ankithans: done

starter/go.mod
namit-chandwani: Yes, but there's still quite some time before we start working on templates, right?
Like we have to refine core and starter then figure out the templating engine to use etc. (Ref: https://github.com/aerogear/charmil/issues/164#issuecomment-883185917)

So maybe till then, we can leave this as a module.
Anyway, the issue of sub-module not working doesn't exist anymore after adding that `replace` directive on line 12, so I don't think keeping this for now will hurt us in any way. WDYT?

Please correct me if I'm wrong

namit-chandwani: Also, we can include a `go.mod` file in the template too. WDYT?

ankithans: done

wtrocki: that will make shell unusable. lets disscus

namit-chandwani: > CLI is gimmick. starter will not be starter if it doest have all files needed to compile

Agreed. 

So I think we need to keep starter and templates as something separate
(Rules out the option 2 here: https://github.com/aerogear/charmil/issues/164#issuecomment-884060416)

PR #158, opened 2021-07-12
feat: set up charmil core features in starter

namit-chandwani:
# Changes made

- Modified `factory` to include config handler and update all examples according to it (Similar to changes made in PR: https://github.com/aerogear/charmil/pull/153).
- Fixed some bugs
- Converted `starter` from a sub-module to a package (by removing `go.mod` and `go.sum` files from the `starter` directory)
- Set up starter CLI to use factory, i18n, logger, etc. from `core` instead of Rhoas SDK
- Set up config management in starter and remove the old implementation files for the same
- Added unit tests for the config package (Will keep improving them, as we proceed)
- Added some inline docs

# Reasons why I converted `starter` from a sub-module to a package

- [Replace directives](https://golang.org/ref/mod#go-mod-file-replace) will be required in a multi-module repository to allow adjacent modules to import each other's code without first publishing and remotely fetching it, which might be difficult to maintain.

- We lose the convenience of being able to run all of the unit tests in the repo for continuous integration. With a single module repository architecture, we can test all services/packages with a single `go test ./…` command, whereas we will require some sort of infrastructure to do the same in a multi-module repository (such as a script to run tests from each directory).

- Cobra uses a single-module repository to achieve something similar to our use case. Link: https://github.com/spf13/cobra/tree/master/cobra

- I believe that our use case does not demand the use of multiple modules, and hence a single module repo with the least amount of maintenance would be a better option for us.

References: https://github.com/golang/go/wiki/Modules#should-i-have-multiple-modules-in-a-single-repository

# Possible Future Changes

#### Didn't make the following changes as a part of this PR, as some of these might require some discussions with the team before proceeding.

- After this PR gets merged, the only import that our starter CLI will be using from RHAOS would be that for generating command docs (`github.com/redhat-developer/app-services-cli/pkg/doc`). Maybe we can add a package with similar functionality to our core.

- Migrate packages like [`build`](https://github.com/aerogear/charmil/tree/main/starter/internal/build) and [`version`](https://github.com/aerogear/charmil/tree/main/starter/pkg/cmd/version) to Charmil Core

- Modify the `load` method of config core to create an empty local config file, if it's not present already.

- Remove unwanted packages like [`starter/pkg/cmd/starter`](https://github.com/aerogear/charmil/tree/main/starter/pkg/cmd/starter) (debatable).

- Include the `GENERATE_DOCS` value inside the starter CLI config struct itself


namit-chandwani:
Mentioning just to avoid any confusion: This PR is ready to review now :)

CC @aerogear/charmil-core

wtrocki:
> Converted starter from a sub-module to a package (by removing go.mod and go.sum files from the starter directory)

That will be problem for user who want to use starter.
In normal scenario they clone and copy folder to use it in their new repo. Now we need extra steps that are not documented or east to figure out.

We need separate go.mod as starter should use released version of the core etc.

wtrocki:
> After this PR gets merged, the only import that our starter CLI will be using from RHAOS would be that for generating command docs (github.com/redhat-developer/app-services-cli/pkg/doc). Maybe we can add a package with similar functionality to our core.

Plugins can use upsteam cobra markdown docs.
I think we should remove this feature as it is very rhoas specific

wtrocki:
> Migrate packages like build and version to Charmil Core

Those are version and docs feature flags. We can kill it. Cobra has its own version command that we can extend (there is issue for it) but my take is to kill that with fire

wtrocki:
@jackdelahunt @ankithans to go with specific review and verification

starter/docs/commands/root_completion.md
```diff
@@ -0,0 +1,26 @@
+## root completion
+
```
ankithans, line 2: do we have completion package in starter or in core till now?

ankithans:
@namit-chandwani how can I run starter CLI?
```bash 
$ go run ./starter/cmd/starter/
2021/07/14 11:06:21 open ./config.json: The system cannot find the file specified.
exit status 1
```
giving me this, after adding config file as well

namit-chandwani:
> @namit-chandwani how can I run starter CLI?
> 
> ```shell
> $ go run ./starter/cmd/starter/
> 2021/07/14 11:06:21 open ./config.json: The system cannot find the file specified.
> exit status 1
> ```
> 
> giving me this, after adding config file as well

Try running these commands:

```console
$ make binary
```
```console
$ ./starter
```

ankithans:
okay this works, but only first time it works.
After first run, it gives
```bash
$ ./starter 
2021/07/14 11:24:35 json: cannot unmarshal object into Go struct field Config.LocConfig.Files of type fs.FS
```

namit-chandwani:
> okay this works, but only first time it works.
> After first run, it gives
> 
> ```shell
> $ ./starter 
> 2021/07/14 11:24:35 json: cannot unmarshal object into Go struct field Config.LocConfig.Files of type fs.FS
> ```

Yes you're right, even I noticed this yesterday and I've added a TODO comment for the same on the code snippet which might be causing that issue ([Link for TODO comment](https://github.com/aerogear/charmil/pull/158/files#diff-178951c5dd5b0ba66c5a5b581a53a1740987a2f0a89ee1145bea3ef1e2593f40R57)).

Working on this right now

namit-chandwani:
# TODO:
- [x] Use cobra to generate command docs instead of the RHOAS package for the same
- [x] Remove build and version packages from starter
- [x] Fix the error: `json: cannot unmarshal object into Go struct field Config.LocConfig.Files of type fs.FS` (Ref: https://github.com/aerogear/charmil/pull/158#issuecomment-879612992)
- [ ] Modify the load method of config core to create an empty local config file, if it's not present already
- [ ] Convert `starter` back to a sub-module
- [ ] Resolve build issues

namit-chandwani:
> > Reasons why I converted
> 
> Those are valid points but my take is that this is tradeoff for now
> Nice that we have compiled list but starter is somewhat separate project - possible separate repo later

@wtrocki If our ultimate aim is to move the starter to a separate repo, isn't now the best time to do that?

I remember that you had asked us about the same before merging that starter PR (Link: https://github.com/aerogear/charmil/pull/145#issuecomment-875534467), but I didn't comment as I wasn't aware of the [drawbacks](https://github.com/aerogear/charmil/pull/158#:~:text=Replace%20directives%20will,from%20each%20directory) of a multi-module repo back then.

I feel that it'll be better if we move the starter to a new repo now itself, to avoid the extra amount of maintenance required by keeping it in the same repo.

WDYT?

namit-chandwani:
@wtrocki Really sorry for following up.

Can you please have a look at the comment just above this one, whenever you're free? :) 

I believe you must have missed it due to many notifications at once. TIA

starter/docs/commands/root_completion.md
namit-chandwani: Yes, if you run the starter CLI you'll notice a command for the same over there.
BTW these docs were autogenerated by Cobra :)

ankithans (approved):
mainly this PR - adds test to config package and integrate it with starter + refactor some parts of starter

namit-chandwani:
I've converted the `starter` dir back to a sub-module for the time being.
I'll create a new discussion for this comment later: https://github.com/aerogear/charmil/pull/158#issuecomment-880089717.

This PR is ready for review and merge now :)
CC @ankithans

PR #159, opened 2021-07-13
feat: example validation

### LINKED ISSUE #116: examples rule validation
Detect invalid examples by parsing command and string in example

Example problem:

https://github.com/redhat-developer/app-services-cli/issues/791

ankithans:
Closes #116 

### Description
- Example Validation Rule
- Tests for rules package

### Type of change
- [x] Documentation change

### Checklist

- [ ] Code Review completed

### Logs
```bash
$ go test ./validator/rules/
commands checked: 16
checks failed: 3
--- FAIL: Test_ExecuteCommand (0.00s)
    cmd_test.go:42: validationErr was not empty, got length: 3; want 0
    cmd_test.go:46: EXAMPLE_MATCHES_RULE: cmd root echo echosub: provided example doesn't match with command
    cmd_test.go:46: EXAMPLE_MATCHES_RULE: cmd root echo times: provided example doesn't match with command
    cmd_test.go:46: EXAMPLE_MATCHES_RULE: cmd root echo: provided example doesn't match with command
FAIL
FAIL    github.com/aerogear/charmil/validator/rules     0.034s
FAIL
```


validator/rules/cmd_test.go
```diff
@@ -0,0 +1,104 @@
...
+				Fields: map[string]bool{"Run": true},
+			},
+			UseMatches: UseMatches{Regexp: `^[^-_+]+$`},
+		},
+	}
+
+	validationErr := ExecuteRules(cmd, &ruleCfg)
+	if len(validationErr) != 0 {
+		t.Errorf("validationErr was not empty, got length: %d; want %d", len(validationErr), 0)
+	}
```
jackdelahunt, line 43: I think we can leave this out, the validation already prints out how many errors occur and we hard code in the want value here anyway.

validator/rules/example_matches.go
```diff
@@ -0,0 +1,45 @@
...
+var ExampleMatchesRule = "EXAMPLE_MATCHES_RULE"
+
+// ExampleMatches defines Regexp to be compared
+type ExampleMatches struct {
+	RuleOptions validator.RuleOptions
+}
+
+// Validate is a method of type Rule Interface
+// which returns validation errors
+// compares the regexp with Use attribute
```
jackdelahunt, line 25: Comment needs to be updated

validator/rules/cmd_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+import (
+	"testing"
+
+	"github.com/aerogear/charmil/validator"
+	"github.com/spf13/cobra"
+)
+
+func Test_ExecuteCommand(t *testing.T) {
+	cmd := rootCmd
```
jackdelahunt, line 11: No need to store this in `cmd`, `rootCmd` is more descriptive anyway

validator/rules/example_matches.go
```diff
@@ -0,0 +1,45 @@
...
+			return errors
+		}
+	}
+
+	cmdPath := cmd.CommandPath()
+	cmdExample := cmd.Example
+
+	// check if cmdPath is not the substring of cmdExample
+	if cmdPath != "" && cmdExample != "" && !strings.Contains(cmdExample, cmdPath) {
+		errors = append(errors, validator.ValidationError{Name: "provided example doesn't match with command", Err: ErrExampleMistmatch, Rule: ExampleMatchesRule, Cmd: cmd})
```
jackdelahunt, line 41: This error message uses the word "match" even though we only check if it contains. So depending on your intended functionality change error message or the condition checked.

Changing the condition will make cases like the echo command fail so maybe that is not what you want.

jackdelahunt (changes requested):

validator/rules/example_matches.go
```diff
@@ -0,0 +1,45 @@
...
+			return errors
+		}
+	}
+
+	cmdPath := cmd.CommandPath()
+	cmdExample := cmd.Example
+
+	// check if cmdPath is not the substring of cmdExample
+	if cmdPath != "" && cmdExample != "" && !strings.Contains(cmdExample, cmdPath) {
+		errors = append(errors, validator.ValidationError{Name: "provided example doesn't correlate with the command", Err: ErrExampleMistmatch, Rule: ExampleMatchesRule, Cmd: cmd})
```
jackdelahunt, line 41: Sorry for being annoying but we need to be a bit more concrete here, so they know what to do to fix this error. 

Maybe something like "provided example doesn't contain command path -> %cmdExample : %cmdPath"

jackdelahunt, line 41: Sorry for being annoying but we need to be a bit more concrete here, so they know what to do to fix this error.

Maybe something like "provided example doesn't contain command path -> %cmdExample : %cmdPath"

jackdelahunt (changes requested):

validator/rules/example_matches.go
ankithans: yep can do this.
> Sorry for being annoying

no issue in that, this helps!

ankithans: > Example:%cmdExample, Path:%cmdPath

we provide cobra cmd object in every error, so this info is avoided in every rule
VlaidationError looks like
```go

Comment (ankithans) [reply validator/rules/example_matches.go:41]: this is actually verbose printing which I missed !

File: validator/example/cmd.go [hunk 6]
```diff
@@ -16,11 +16,19 @@ type command struct {
 
 func NewCommand() *cobra.Command {
 
+	cmd100 := &cobra.Command{
+		Use:     "cmd100",
```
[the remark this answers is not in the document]
ankithans: shall we remove examples directory, as we have tests for validator now?

PR #161, opened 2021-07-14
fix: skip validation for children

### LINKED ISSUE #151: SkipChildren option not working in validator

ankithans:
Closes #151 

### Description
- fix skip children commands by specifying path

### Type of change

### DIFF
--- a/validator/rules/cmd_test.go
+++ b/validator/rules/cmd_test.go
@@ -3,7 +3,6 @@ package rules
 import (
 	"testing"
 
-	"github.com/aerogear/charmil/validator"
 	"github.com/spf13/cobra"
 )
 
@@ -13,14 +12,12 @@ func Test_ExecuteCommand(t *testing.T) {
 	// default config can also be overrided
 	ruleCfg := ValidatorConfig{
 		ValidatorOptions: ValidatorOptions{
-			SkipCommands: map[string]bool{"cmd100 cmd0 subcmd01": true},
-			Verbose:      true,
+			SkipChildren: map[string]bool{
+				"root echo": true,
+			},
 		},
 		ValidatorRules: ValidatorRules{
 			Length: Length{
-				RuleOptions: validator.RuleOptions{
-					SkipCommands: map[string]bool{"cmd100": true},
-				},
 				Limits: map[string]Limit{
 					"Use":     {Min: 1},
 					"Example": {Min: 3},
@@ -28,9 +25,6 @@ func Test_ExecuteCommand(t *testing.T) {
 				},
 			},
 			MustExist: MustExist{
-				RuleOptions: validator.RuleOptions{
-					SkipCommands: map[string]bool{"cmd100": true},
-				},
 				Fields: map[string]bool{"Run": true},
 			},
 			UseMatches: UseMatches{Regexp: `^[^-_+]+$`},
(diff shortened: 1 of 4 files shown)


jjkiely:
Have you verified that this fixes the issue? If not can you provide some steps I can follow in order to do so? Thanks!

validator/rules/cmd_test.go
```diff
@@ -13,24 +12,19 @@ func Test_ExecuteCommand(t *testing.T) {
 	// default config can also be overrided
 	ruleCfg := ValidatorConfig{
 		ValidatorOptions: ValidatorOptions{
-			SkipCommands: map[string]bool{"cmd100 cmd0 subcmd01": true},
```
wtrocki, line 16: Is this removed?

ankithans:
yep verified myself.
steps are:
- run `make test/validator` or `go run ./validator/rules`
- change the stuff in `validator/rules/cmd_test.go` 

```go
ValidatorOptions: ValidatorOptions{
	SkipChildren: map[string]bool{
		"root echo": true,
	},
}
```
this will stop checking root echo and root echo .... children

validator/rules/cmd_test.go
ankithans: no

ankithans: just removed from test

wtrocki:
This is breaking change also removes other flags that we have documented. Best to put more info into PR.

I would like if we can avoid having multiple flags for skipping

Skip for children can be done by 
```
CMD data manipulators *
```

Where last `*` means children only.
I'm little bit like broken record but..

lets start with docs and api description

ankithans:
> Skip for children can be done by
CMD data manipulators *

got it. this looks cool and is easy to use. Let's do this.

> I'm little bit like broken record but..
> lets start with docs and api description

you telling about skipping issue or the whole project in general

wtrocki:
> you telling about skipping issue or the whole project in general.

No. I meant that I will not review PRs without docs :)

ankithans:
oh okay, if you haven't noticed, this was not a breaking change and docs was already there : https://github.com/aerogear/charmil/blob/validator-skip/docs/src/validator.md#ignore-commands

but now let's do the proposed change and will update the docs.

ankithans:
docs: https://github.com/aerogear/charmil/blob/validator-skip/docs/src/validator.md#ignore-commands

ankithans:
yep, we have a test using SkipCommands attribute

validator/rules/executor.go
```diff
@@ -70,6 +72,14 @@ func executeRecursive(cmd *cobra.Command, info *validator.StatusLog, ruleConfig
 
 // executeRulesChildren execute rules on children of cmd
 func executeRulesChildren(cmd *cobra.Command, info *validator.StatusLog, ruleConfig *RuleConfig, userValidatorConfig *ValidatorConfig) []validator.ValidationError {
+
+	// if command's children needs to be ignored
+	if val, ok := userValidatorConfig.ValidatorOptions.SkipCommands[cmd.CommandPath()+"*"]; ok {
```
wtrocki, line 77: This should be helper method

wtrocki (approved):

wtrocki:
> but now let's do the proposed change and will update the docs.

Docs first please.

ankithans:
> Docs first please next time.

yess 💪

validator/rules/executor.go
ankithans: done

wtrocki:
We can release now

PR #171, opened 2021-07-22
feat: add CRUD files generation functionality to charmil CLI

### LINKED ISSUE #67: [Wrapper] Rapid Creation of CRUD command line tools.
In some situations developers can avoid a lots of boilerplate if we enable them to have 

Imagine that you have CLI like RHOAS CLI - where there will be multiple services performing typical crud operations:

```
[wtrocki@graphapi app-services-cli (srs-management-cli ✗)]$ rhoas kafka
Create, view, use, and manage your Apache Kafka instances

Usage:
  rhoas kafka [command]

Available Commands:
  create        Create an Apache Kafka instance
  delete        Delete an Apache Kafka instance
  describe      View configuration details of an Apache Kafka instance
  list          List all Apache Kafka instances
  use           Set the current Apache Kafka instance
  
```

Adding new service usually means duplication of the commands, error handling etc.
We should research if we can provide some clever plugin (still using charmil core factory)

namit-chandwani:
Closes #67 

### Description
- Added CRUD files generation functionality to Charmil CLI by adding a sub-command named `crud`.
- With the help of this `crud` command, developers will be able to eliminate a lot of boilerplate, in CLIs containing multiple services that perform standard CRUD operations.

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

# TODO:
- [ ] Think of a way to generate locale file for the generated CRUD files (Blocked by the implementation of `charmil init` command. Can't be resolved at this moment)
- [x] Add inline docs
- [x] Add dev usage docs


pkg/cmd/crud/crud.go
```diff
@@ -0,0 +1,76 @@
...
+	tmpl "github.com/aerogear/charmil/pkg/template"
+	"github.com/spf13/cobra"
+)
+
+type FlagVariables struct {
+	path string
+
+	Singular string
+	Plural   string
+}
```
ankithans, line 18: shall we use this github.com/manifoldco/promptui? Instead of flags. To make it cooler!
whatever feels good. Need suggestion here @wtrocki

wtrocki: Flags and interactive mode can be supported.
Flags are good start

namit-chandwani: > shall we use this github.com/manifoldco/promptui

I had a look at its readme just now, looks really cool.

namit-chandwani:
This PR is now ready for review :)
CC @aerogear/charmil

pkg/template/crud.go
```diff
@@ -0,0 +1,210 @@
+package template
+
+// CreateCrudTemplate returns the template of the `create` file as a byte array
+func CreateCrudTemplate() []byte {
+	return []byte(`// Create file for {{.Singular}} instances
+
+package crud
+
```
wtrocki, line 8: golang has embedded template engine.
Alternatively go embed can use actual source code as strings.
I like go embed as template could compile etc.

pkg/template/crud.go
```diff
@@ -0,0 +1,210 @@
...
+			return runUse(opts, f)
+		},
+	}
+
+	return cmd
+}
+
+func runUse(opts *useOptions, f *factory.Factory) error {
+	// Add your implementation here
+
```
wtrocki, line 206: I would split those into separate files if possible

pkg/template/crud.go
namit-chandwani: Resolved via commit: https://github.com/aerogear/charmil/pull/171/commits/0972c08917e0ea8019b80fb58c21a4d146eaa307

wtrocki (approved):
This PR proves that giving extra chances really makes difference. 
Amazing work overall. CRUD commands need more work - flags, validation etc. that we could extend in future PR's this is good base to start. 

Did not verified it

wtrocki:
Leavin @ankithans to make final call

namit-chandwani:
> go:embed works with folders as well.
> It will be less repetition this way.

Agreed. I'm making this change right now

pkg/template/crud/create.go
```diff
@@ -0,0 +1,36 @@
+// Create file for {{.Singular}} instances
```
ankithans, line 1: one optional change. We can have this as one template, instead of create, delete, describe, list, use as different templates. As just names are changed and rest all is repeated. I think this can be done by making a map of names and a loop. @namit-chandwani @wtrocki 

Overall works as expected 🤞

namit-chandwani: Good catch!
Initially, I had assumed that all CRUD files would contain something unique, which is why I had started with separate files.
Since that is not the case anymore, I'll change it now :)

wtrocki: -1000000 please not do it. This is optimalization based on current state. We will add specific code to it later

pkg/template/crud/list.go
```diff
@@ -0,0 +1,37 @@
...
+	// Add your option fields here
+
+}
+
+// NewListCommand creates a new command for listing instances.
+func NewListCommand(f *factory.Factory) *cobra.Command {
+	opts := &listOptions{}
+
+	cmd := &cobra.Command{
+		Use:     f.Localizer.LocalizeByID("crud.cmd.list.use"),
```
wtrocki, line 20: are we creating those localizations?

pkg/template/crud/create.go
wtrocki: Actually we need common flags without them commands will not work

wtrocki:
as follow up we need flags validation etc. Maybe we can mount some core stuff like config to use etc.

wtrocki (approved):

pkg/template/crud/list.go
namit-chandwani: Yes, I had added a locale file for this initially but then removed it later through a different commit.
As mentioned [over here](https://github.com/aerogear/charmil/pull/171#:~:text=Think%20of%20a%20way%20to%20generate%20locale%20file%20for%20the%20generated%20CRUD%20files%20(Blocked%20by%20the%20implementation%20of%20charmil%20init%20command.%20Can%27t%20be%20resolved%20at%20this%20moment)), the location where the locale file should be generated, depends upon the `charmil init` command implementation.

I'm creating an issue for this and will resolve it through a separate PR once the `charmil init` part is completed.

wtrocki: Lets add that. Wuthout it there is not much value in those

wtrocki: my prefference will be to have single yaml file

ankithans: find the locales directory and add locales there.
It will be present in /cmd/cliname/locales/....

namit-chandwani: > find the locales directory and add locales there.
> In a cli, It will be present in /cmd/cliname/locales/....

Yes, this is exactly what I wanted to know about the `charmil init` implementation. Will add locale file according to this.
Thanks :)

PR #173, opened 2021-07-25
chore: remove starter from main

ankithans:
remove starter from main. No need to maintain the starter at 2 locations - in main and in separate repo [here](https://github.com/ankithans/charmil-starter-template)


wtrocki:
but we do not have second location :)
You mean your experiment in your own org?

wtrocki (changes requested):
Let me check why we do it tommorow

ankithans:
yep sure! we can move it under aerogear or any other solution..

wtrocki:
I do not think we should remove starter for now only because we will lose ability to work with unreleased changes of the core, validator etc. Separate repository will be painful to work with. 

We can always create separate repo for this no problem. Trick is that we do not have much content and there is no reason for us to slow things down. If you worried about init not being able to clone/extract subfolder let's create separate repository that we can manually sync with the main on release - that would also be great way to separate main from released versions used for init. WDYT?

ankithans:
> let's create separate repository that we can manually sync with the main on release - that would also be great way to separate main from released versions used for init. WDYT?

this looks great!

wtrocki:
https://github.com/aerogear/charmil-starter/settings/access

PR #188, opened 2021-07-28
chore: refactor repo structure

### LINKED ISSUE #180: Invalid repository structure
### Feature or problem description

We currently have number of components in the repo and using 2 mixed patterns:

pkg for root where we have CLI commands

specific folders like core.
I think pkg doesn't make sense in this case and we should use submodules for each component:

- Validator
- Core

We can use pkg and internal folders within those but that is not required.

ankithans:
Closes #180 
### Description
rm pkg, put pkg into core


wtrocki:
Before applying changes put comment what you planning to do to get validation. It saves time :D

wtrocki:
Changes make no sense for me. 

Scafolders are part of the CLI not core.

ankithans:
so we can move it to cmd folder(in the root). I followed pkg, just for the sake - we provide starter and we should follow it.

wtrocki:
We should have functional folders for each component. cmd/pkg/internal are not folders. they have specific meaning and should not be mixed together. 

That is classical project layout: https://github.com/golang-standards/project-layout

However it will not work for our use case as we keep multiple elements in the repo.

wtrocki:
Side note. Starter will be the only module that we will keep as we do not need it to be consumable as lib, but it will be used by init etc. Rest will be simply folders.

ankithans:
> Let's document repository structure

yes...

validator, starter, core - already a folder
cli - present in cmd/ folder. We don't need pkg folder and we can put comands and templates in cmd/ folder.

the above change will be fine?

wtrocki:
Please read: https://github.com/golang-standards/project-layout

cmd folder is special folder. Should not contain much code.

ankithans:
yeah sorry, code should not be used or imported into other projects, so need to put that into internal

wtrocki:
Agreed

ankithans:
> cli/pkg and cli/cmd?

yeah
```
cli
   cmd
   pkg
core
starter
validator
```
can we rename cli to charmil? (this can be ignored)

namit-chandwani:
>    pkg - here pkg or internal??

Internal would be better IMO, since the code in `cli/pkg` won't be of any use outside this repo

ankithans:
@wtrocki ready for review

CONTRIBUTING.md
```diff
@@ -9,6 +9,14 @@ The following should be to be installed on your device in order to contribute to
...
+### Folder Structure
+Charmil repository contains a number of components that can be consumed independently. In the root directory, the following folders can be found:
+
+- **cli** consists of command line tool for using charmil.
+- **core** consists of charmil SDK 
+- **starter** is a charmil starter template project
+- **validator** is the testing library to test cobra commands
+
 ### Running Charmil CLI Commands
 You can run the CLI commands files directly with go run. All commands and subcommands are in the ./pkg/cmd folder.
```
namit-chandwani, line 21: Maybe you can change this too now

```suggestion
You can run the CLI commands files directly with go run. All commands and subcommands are in the ./cli/internal/cmd folder.
```

namit-chandwani (changes requested):

wtrocki (approved):

PR #195, opened 2021-07-30
chore: move factory to starter

### LINKED ISSUE #162: Move Factory from core to starter
Plugin needs it's own factory while development, and at the time of dev, we cannot pass host's factory instance to plugin. There can be these solutions:
- have different factories, no connections
- have different factories, plugin's factory to be override by host's factory instance automatically(essential attributes)

We may like to think a approach that will work for config management between plugin & host.

ankithans:
Closes #162 

### Description
Move factory from core to starter.


ankithans:
need some initial reviews. Not sure we are doing it right! As I see, just the boiler plate increases in starter 😅

ankithans:
and there is this file `.DS_Store` in main, not present in my IDE. Don't know from where it came.

wtrocki:
Git ignore .DS_Store

wtrocki:
@aerogear/charmil

wtrocki:
Some strange files get into this PR but path is good.

starter/docs/commands/root_completion_bash.md
```diff
@@ -4,7 +4,7 @@ Generate command completion script for Bash shell
 
 ### Synopsis
 
-' Generate command completion script for Bash shell. Installing on Linux: 1. Run `rhoas completion bash > rhoas_completions` to create the script file. 2. Run `sudo mv rhoas_completions /etc/bash_completion.d/rhoas` to place the script in a special Bash completions directory. Installing on macOS: 1. Run `rhoas completion bash > rhoas_completions` to create the script file. 2. Run `sudo mv rhoas_completions /usr/local/etc/bash_completion.d/rhoas` to place the script in a special Bash completions directory. '
+' Generate command completion script for Bash shell. Installing on Linux: 1. Run `<no value> completion bash > <no value>_completions` to create the script file. 2. Run `sudo mv <no value>_completions /etc/bash_completion.d/<no value>` to place the script in a special Bash completions directory. Installing on macOS: 1. Run `<no value> completion bash > <no value>_completions` to create the script file. 2. Run `sudo mv <no value>_completions /usr/local/etc/bash_completion.d/<no value>` to place the script in a special Bash completions directory. '
```
namit-chandwani, line 7: I think the `{{ .CliName }}` placeholder is being replaced by `<no value>` values over here, since `CliName` var is empty currently.
So maybe you can add a default value to it, to fix it temporarily.

WDYT?

ankithans: thanks, I didn't noticed this. Will fix it.

ankithans:
@wtrocki what we can do more here?

wtrocki:
merge

PR #233, opened 2021-08-17
fix(core): improve config management package

### LINKED ISSUE #184: Core config feature is not working as expected.
### Feature or problem description

I have been playing with the config and it looks like it is not working as it should.

Config is not properly strongly typed:

```
	cfg interface{}
```

Config should provide interface for load and save where save and load triggered from the plugin will actually execute top level save and load functions and return proper results. Go doesn't support generics but what we really need to is to bridge specific load and save functions in plugin with host cli. If we can override plugin save methods and load methods on the host that will be all that is needed.

The current implementation is crashing for me when setting up in root and plugin with sample config on MergePluginCfg.

namit-chandwani:
# WIP (Up for an early discussion)

Closes #184 

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/core/newconfig/config.go
+++ b/core/newconfig/config.go
@@ -1,0 +1,100 @@
+package newconfig
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/tidwall/gjson"
+	"github.com/tidwall/sjson"
+)
+
+const envName = "CHARMIL_CONFIG"
+
+// TODO: Temporary struct. Find a way to make things work while keeping the config strongly-typed
+type Config struct {
+}
+
+type IConfig interface {
+	Load() (*Config, error)
+	Save(cfg *Config) error
+	Remove() error
+	Location() (string, error)
+}
+
+func NewCfgHandler() IConfig {
+	// TODO: Find a way to dynamically return a config handler from the 2 possible options (ie. hostCfgHandler or pluginCfgHandler).
+	// aka Strategy pattern
+
+	cfg := &hostCfgHandler{}
+	// cfg := &pluginCfgHandler{}
+
+	return cfg
+}
+
+func hasCustomLocation() bool {
+	cfg := os.Getenv(envName)
+	return cfg != ""
+}
+
+func defaultDir() (string, error) {
+	userCfgDir, err := os.UserConfigDir()
+	if err != nil {
+		return "", err
+	}
+
+	return filepath.Join(userCfgDir, "charmil"), nil
+}
+
+// readFile reads the file specified by filePath and returns its contents.
+func readFile(filePath string) ([]byte, error) {
+	_, err := os.Stat(filePath)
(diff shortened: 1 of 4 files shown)


core/newconfig/config.go
```diff
@@ -0,0 +1,100 @@
+package newconfig
```
wtrocki, line 1: new?

core/newconfig/config.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/tidwall/sjson"
+)
+
+const envName = "CHARMIL_CONFIG"
+
+// TODO: Temporary struct. Find a way to make things work while keeping the config strongly-typed
+type Config struct {
+}
+
+type IConfig interface {
```
wtrocki, line 19: Too generic name

core/newconfig/config.go
namit-chandwani: Just did it temporarily so that you can notice the changes easily.
Will replace it with the old config package once everything here is done

core/newconfig/config.go
```diff
@@ -0,0 +1,100 @@
...
+func writeFile(filePath string, data []byte) error {
+	err := ioutil.WriteFile(filePath, data, 0600)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getJsonSubkey(data []byte, keyName string) []byte {
```
wtrocki, line 78: Where is this used?

core/newconfig/host.go
```diff
@@ -0,0 +1,85 @@
...
+
+	err = writeFile(f, data)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (h *hostCfgHandler) Remove() error {
```
wtrocki, line 51: What is the purpose of remove?

core/newconfig/config.go
```diff
@@ -0,0 +1,100 @@
...
+func setJsonSubkey(data []byte, keyName string, cfg *Config) ([]byte, error) {
+	val, err := sjson.Set(string(data), keyName, cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	return []byte(val), nil
+}
+
+func deleteJsonSubkey(data []byte, keyName string) ([]byte, error) {
```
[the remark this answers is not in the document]
namit-chandwani: Used over here: https://github.com/aerogear/charmil/pull/233/files#diff-f2c6629fa77680e59b436813801be8f5196433e238ded04832196387a36905a6R72

wtrocki:
@namit-chandwani I do not see how this would work.

Approach with subkeys is not going to allow us to load and write structure. Please look at example plugin and example host. Modify them to use **different** configs and we can discuss later hoe to swap the config from host. Proposed PR doesnt differ much with config on master

wtrocki:
Ignore comments. Lets work with @ankithans to help him with examples and solve config outside core first

